PR #20, opened 2023-10-13
applet example

GabrielaReyna:


nanaian (approved):
LGTM- is this PR meant to be labeled as draft?
Let's make sure we update the examples' script tag URLs to the release candidate one once it's out tomorrow

PR #29, opened 2023-11-02
blog header doodle

GabrielaReyna:


nanaian (changes requested):
Can you place blog-header.png in `src/pages/blog` and use [Image](https://docs.astro.build/en/guides/images/#image--astroassets) from `astro:assets` instead of an img tag to optimise the image?

GabrielaReyna:
Image as Astro asset lowers the quality considerably.  It looks slightly blurry, which is not too bad but I can tell the difference.

nanaian (approved):

PR #38, opened 2023-11-18
fixup! typo

cattabiani:
This made the call print always the rectangle as black


nanaian (approved):
Good spot, thanks again 😁

PR #39, opened 2023-11-18
bound platform movement

cattabiani:
Atm the platform can move out of the canvas. Let's add some bounds! 

Also if someone wants to tell me why there is fmin and fmax (I thought they were for floats) but not min/max I would gladly add this to the docs too

EDIT: I also increased the speed of the platform. I cannot keep up with the ball!


nanaian (commented):
Good idea bounding movement, and I agree with increasing the speed. We definitely need to explain why we're using `fmin`/`fmax` over the STL functions.. did `std::min`/`std::max` not work? Perhaps @yuri91 can provide some insight here

yuri91:
What is the error that is shown if std::min is used?

cattabiani:
Atm I moved on and cannot try with the exact configuration I had before. I tried to add it to my current project and it is working. I amso checked the inputs I was using and they were all ints. I can tell you only that it could not find the function and suggested to use fmin/fmax instead. Since I just wanted to see if it was working, I moved on without recording the error.

cattabiani:
added the discord explanation in the file.

nanaian (changes requested):
Thanks guys. A small edit then LGTM. Also please run `pnpm format` so that CI passes :)

PR #48, opened 2023-12-13
file system guide improved

GabrielaReyna:
Pass on the filesystem guide:
- Redaction improvements
- examples for each mount-point
- Info rearrangements


src/content/docs/cheerpj3/11-guides/File-System-support.md
```diff
@@ -2,64 +2,96 @@
...
+
+> [!warning]Important
+> CheerpJ provides access to a virtualized filesystem, which does not correspond to the local computer. Accessing local files from the browser is forbidden due to security reasons.
+
 ## `/app/` mount point
 
 The /app/ mount point corresponds to a virtual read-only, HTTP-based filesystem. `/app/` is used to access JAR files and data from your local server.
 
-The `/app/` directory is virtual, and only exists inside of CheerpJ. It is required to distinguish files from the local server from runtime files and files stored in the browser database.
+This mount point is virtual and only exists inside of CheerpJ. It is required to distinguish between local server, runtime files and files stored in the browser database.
```
alexp-sssup, line 26: I am not sure about the purpose of this sentence. All mount points are virtual and only exists inside of CheerpJ

src/content/docs/cheerpj3/11-guides/File-System-support.md
```diff
@@ -2,64 +2,96 @@
...
+- `/app/subdirectory/example.txt` would be the same as `http://127.0.0.1:8080/subdirectory/example.txt`
+
+Considering the examples above, to run a JAR with [`cheerpjRunJar`] and assuming it is stored in the root of the web server, it should be done as follows:
+
+```js
+cheerpjRunJar("/app/my_application_archive.jar");
+```
+
+> [!tip] Storing JAR files
+> The /app/ mount point is the most common location to store the application's JARs but this is not necessarily mandatory.
```
alexp-sssup, line 41: "necessarily mandatory" -> redundant?

src/content/docs/cheerpj3/11-guides/File-System-support.md
```diff
@@ -2,64 +2,96 @@
...
 
 ## `/files/` mount point
 
-The `/files/` mount point corresponds to a virtual read-write, IndexedDB-based filesystem. `/files/` is used to store persistent data on the browser client.
+The `/files/` mount point corresponds to a virtual read-write, IndexedDB-based filesystem and it is used to store persistent data in the browser client.
+
+Writing files into the `/files/` mount point is only possible from inside the Java application and specifying the root of the file path explicitly. For example:
```
alexp-sssup, line 47: > and specifying the root of the file path explicitly

This is the case for any of  backend. There is no reason why `/files` should be special.

src/content/docs/cheerpj3/11-guides/File-System-support.md
```diff
@@ -2,64 +2,96 @@
...
+```java
+File file = new File("/files/myfile.ext");
+OutputStream out = new FileOutputStream(file);
+out.close();
+```
+
+To read files from the `/files/` mount-point using JavaScript is done with the [`cjFileBlob`] API as follows:
+
+```js
+const blob = await cjFileBlob("/files/myfile.ext");
```
alexp-sssup, line 58: Although this is currently broken this API should work in general for 3.0 final release. It's not intended to be specific to /files/

src/content/docs/cheerpj3/11-guides/File-System-support.md
nanaian: I would remove this para entirely

nanaian (approved):

PR #50, opened 2023-12-20
networking guide first iteration

GabrielaReyna:
This first version is mainly information gathered from different channels (discord conversations, blog posts etc).
I will be iterating on top of this with use-case examples. The aim is to have some information out there. 

Content:
- Explanations on same-origin HTTP(S) requests and generalized networking
- How to connect to a Tailscale network with CheerpJ
- Suggestion on using LT Headscale fork if self-hosting


src/content/docs/cheerpj3/11-guides/Networking.md
```diff
@@ -0,0 +1,87 @@
...
+CheerpJ 3.0 supports different networking scenarios that are divided into two categories:
+
+- Same-origin HTTP/HTTPS requests using fetch seamlessly.
+- Generalized networking via Tailscale.
+
+## Same origin HTTP/HTTPS requests
+
+A Java application running in the browser with CheerpJ can request resources to the local server (same origin) via fetch through the HTTP/HTTPS protocols. As expected, these requests are asynchronous. Remember that for a request to be considered same-origin, it needs to match the scheme, hostname and port between the requester and the receiver.
+
+The CheerpJ runtime API implements its own HTTP/HTTPS handler using the browser `fetch()`. To use this handler you need to pass into `cheerpjInit` with the `fetch` option as shown below:
```
alexp-sssup, line 15: That is not actually required, we allowed fetch to be overridden for JNLP impl, but we should not document that.

By default the normal fetch is used anyway

src/content/docs/cheerpj3/11-guides/Networking.md
```diff
@@ -0,0 +1,87 @@
...
+	});
+}
+cheerpjInit({ fetch: fetchExample });
+```
+
+## Generalized networking
+
+When it comes to generalized networking such as access to private network services, peer-to-peer connections or access to the wider internet, there are some browser-enforced limitations to enhance user's security. These limitations are bound to the browser not exposing access to lower-level protocols such as UDP and TCP. The browser only allows one to perform HTTP(S) requests ruled by CORS policies.
+
+Another important scenario is the usage of WebSockets, where there is a two-way interaction between the user and the server in an event-driven manner. A WebSocket connection starts with a request from the browser via HTTP that when accepted, this connection is upgraded and does not conform to the HTTP protocol anymore. This way the user and the server can keep a connection and send bidirectional messages until this connection is closed from one of the endpoints.
```
alexp-sssup, line 30: These 2 paragraphs add confusion in my opinion

src/content/docs/cheerpj3/11-guides/Networking.md
```diff
@@ -0,0 +1,87 @@
...
+#### Connecting your application to a Tailscale network
+
+Connecting the Java application client to your Tailscale network is as simple as providing client authentication via `cheerpjInit()`:
+
+Example for pre-authenticated nodes:
+
+```js
+cheerpjInit({
+	tailscaleControlUrl: "https://my.url.com/",
+	tailscaleDnsUrl: "1.1.1.1",
```
alexp-sssup, line 53: We use 8.8.8.8 by default, so you can avoid this

src/content/docs/cheerpj3/11-guides/Networking.md
GabrielaReyna: You're right, I have re-structured this information.

src/content/docs/cheerpj3/11-guides/Networking.md
```diff
@@ -0,0 +1,77 @@
...
+```js
+cheerpjInit({
+	tailscaleControlUrl: "https://my.url.com/",
+	tailscaleAuthKey: "AuthKeyStringGoesHere",
+});
+```
+
+What is happening here?
+
+- `tailscaleControlUrl` is a string URL of the Tailscale control plane which verifies the user's identity. Only pass this option if you are [self-hosting Tailscale](https://github.com/leaningtech/headscale)
```
nanaian, line 52: missing "."

src/content/docs/cheerpj3/11-guides/Networking.md
```diff
@@ -0,0 +1,77 @@
...
+- `tailscaleAuthKey` is string with an auth key to register new nodes that are pre-authenticated. You can create an auth key [here](https://login.tailscale.com/admin/settings/keys).
+
+Example to prompt the user for manual login:
+
+```js
+cheerpjInit({
+	tailscaleControlUrl: "https://my.url.com/",
+	tailscaleLoginUrlCb: function (url) {
+		// your function code here to continue with login
```
nanaian, line 61: An actual example here that returns something would be good. Maybe see webvm for an example

src/content/docs/cheerpj3/11-guides/Networking.md
nanaian, line 52: Link to the #self-hosting-headscale section I think

src/content/docs/cheerpj3/11-guides/Networking.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Networking
+description: Networking with CheerpJ
+---
+
+CheerpJ 3.0 supports different networking scenarios that are divided into two categories:
+
+- Same-origin HTTP/HTTPS requests using fetch seamlessly.
+- Generalized networking via Tailscale.
+
```
nanaian, line 10: IMO it's worth making it explicit that you must use the latter for anything that isn't HTTP(S), such as opening TCP or UDP sockets.

src/content/docs/cheerpj3/11-guides/Networking.md
GabrielaReyna: I think tailscaleLoginUrlCb does not return anything, but I added a few lines on how it would be to prompt the login to a different tab for now (based on the WebVM example). Will add something more complex as I play with it.

nanaian: My bad. Thanks, lgtm

nanaian (approved):

src/content/docs/cheerpj3/10-getting-started/02-Java-library.md
```diff
@@ -16,6 +16,9 @@ await cheerpjInit();
 const cj = await cheerpjRunLibrary("/app/library.jar");
 ```
 
+> [!warning]Important
```
nanaian, line 19: "

PR #53, opened 2023-12-29
Update code styling and fix BlogPostCard on mobile

nanaian:
Updates to the code styling, including:
- New code font
- Nicer code block theme
- `inline code` background

Also fixes an issue with the BlogPostCard component on small widths.


nanaian:
Before:
image

After:
image

PR #54, opened 2024-01-03
Fix command being cut off short

Hyxogen:
Previously the command could be cut off short with no easy way to see the rest of it.

Fixed by placing the command in a code block which allows horizontal scrolling.

Before on my machine:
image


nanaian:
Thanks for the fix! This loses the emboldening of `-cheerp-make-module=es6` - I'll add that with [expressive-code syntax](https://github.com/expressive-code/expressive-code/blob/main/packages/@expressive-code/plugin-text-markers/README.md#usage-in-markdown--mdx-documents) before merging

PR #63, opened 2024-01-24
How does CheerpJ works? draft

GabrielaReyna:
Explanation page


src/content/docs/cheerpj3/11-guides/CheerpJ-explanation.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: CheerpJ components
+description: How does CheerpJ works?
```
nanaian, line 3: This should be "how does CheerpJ work?" but I think we can do better since it's obvious this page is about CheerpJ.

I'd suggest (in order of my preference):

- Architecture
- How it works
- Technical details

src/content/docs/cheerpj3/11-guides/CheerpJ-explanation.mdx
```diff
@@ -0,0 +1,81 @@
+---
```
nanaian, line 1: I don't like how 'CheerpJ' is in the file path / url of this page, but "explanation" on its own is too vague. Also `guides` is not the right folder for this.

Maybe `14-explanation/architecture.mdx`?

src/content/docs/cheerpj3/11-guides/CheerpJ-explanation.mdx
```diff
@@ -0,0 +1,81 @@
...
+## Use cases
+
+Here we provide a list of the most common uses-cases that CheerpJ supports:
+
+- Restoration of legacy gaming, educational or personal content.
+- Modernization of enterprise software such Java applications, applets, Web Start / JNLP.
+- Building web applications with Java libraries.
+- Fully client-side Java development environments and playgrounds.
+
+CheerpJ is free for personal use and technical evaluation. See [licensing] for details.
```
nanaian, line 71: This doesn't feel like it fits here, it's not really an explanation. Maybe a whole separate use cases page where we go into more detail? We could also talk about actual clients (linking to the [case studies](https://leaningtech.com/case-studies/)) and migrate the community project list to this page too

src/content/docs/cheerpj3/11-guides/CheerpJ-explanation.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+Swing applications, will render exactly as they do in native. Swing Look&Feel is also supported, including 3rd party ones. Multi-window applications are supported, with keyboard focus being managed as expected. Integration with the system clipboard can be enabled via an initialization option.
+
+### The virtualized file system
+
+CheepJ provides [multiple filesystem backends](cheerpj3/guides/File-System-support) to accommodate different application needs, including access to server-hosted files and persistent local storage.
+
+![](/cheerpj2/assets/filesystem.png)
+
+You can learn more about CheerpJ's filesystem at our [files and filesystems guide].
```
nanaian, line 54: s/at/in

I think this link, and the link in the networking section, would be better off as LinkButtons with a right arrow icon

src/content/docs/cheerpj3/11-guides/CheerpJ-explanation.mdx
```diff
@@ -0,0 +1,81 @@
...
+- Modernization of enterprise software such Java applications, applets, Web Start / JNLP.
+- Building web applications with Java libraries.
+- Fully client-side Java development environments and playgrounds.
+
+CheerpJ is free for personal use and technical evaluation. See [licensing] for details.
+
+## Further reading
+
+- [Getting started](/cheerpj3/getting-started)
+- [Take a look to our tutorials](/cheerpj3/examples)
```
nanaian, line 76: Not sure we need this

PR #70, opened 2024-02-27
CheerpJ JNLP runner page

GabrielaReyna:


src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+			<div>
+				<h1
+					class="text-white text-4xl md:text-5xl lg:text-6xl xl:text-7xl font-semibold"
+				>
+					Run
+					<span class="font-bold text-primary-500"> Java Web Start</span>
+					<br />apps in the browser
+				</h1>
+				<p class="text-xl mt-8 text-stone-400 leading-6">
+					Unleash the power of CheerpJ now available in a easy-to-use browser
```
alexp-sssup, line 28: CheerpJ `,` now available

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+				The <span class="font-bold text-primary-500"> JNLP</span> browser
+				extension
+			</h3>
+			<p class="text-right">
+				Follow our channels to stay up to date with our community
+			</p>
+			<div class="mt-8 flex flex-wrap justify-end gap-4 pt-10">
+				<LinkButton
+					href="https://github.com/leaningtech/cheerpj-jnlprunner"
+					label="Repository"
```
alexp-sssup, line 173: Hmmm.. "Bug reports"? "Open Issues"?

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+		<div class="flex items-center col-span-2">
+			<div>
+				<h1
+					class="text-white text-4xl md:text-5xl lg:text-6xl xl:text-7xl font-semibold"
+				>
+					Run
+					<span class="font-bold text-primary-500"> Java Web Start</span>
+					<br />apps in the browser
+				</h1>
+				<p class="text-xl mt-8 text-stone-400 leading-6">
```
bates64, line 25: Can you increase the leading

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+					Unleash the power of CheerpJ now available in a easy-to-use browser
+					extension. Run Java Web Start Applications, including Oracle
+					E-Business Suite and Oracle Forms Applications, without a Java Desktop
+					installation.
```
bates64: Maybe have the second sentence in its own paragraph? It's a fairly big block of 3 lines of text at the moment.

That is

```
Unleash the power of CheerpJ, now available in a easy-to-use browser extension!

Run Java Web Start Applications, including Oracle E-Business Suite and
Oracle Forms Applications, without a Java Desktop installation.
```

If it looks OK, we should also consider using non breaking spaces (`&nbsp;`) instead of normal spaces in phrases like "Java Web Start Applications". This will prevent reflowing like

```
Run Java Web Start Applications, including Oracle
E-Business Suite and Oracle Forms
Applications, without a Java
Desktop installation.
```

which is hard to parse imo.

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+					<LinkButton
+						href="#ecosystem"
+						label="Add to Chrome"
+						type="primary"
+						iconRight="ri:chrome-fill"
+					/>
+					<LinkButton
+						href="#"
+						label="Add to Edge"
+						type="secondary"
```
bates64, line 40: Having this secondary doesn't make sense to me.

Can we instead make a new component that picks the right LinkButton to use/highlight depending on what the user's browser is? Instead of user agent sniffing (to be deprecated), we can use [User Agent Client Hints](https://developer.chrome.com/docs/privacy-security/user-agent-client-hints) on Chromium derivatives like Chrome and Edge (that is, the API is not supported on Firefox at the moment, but we don't care).

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+				</h1>
+				<p class="text-xl mt-8 text-stone-400 leading-6">
+					Unleash the power of CheerpJ now available in a easy-to-use browser
+					extension. Run Java Web Start Applications, including Oracle
+					E-Business Suite and Oracle Forms Applications, without a Java Desktop
+					installation.
+				</p>
+				<div class="mt-8 flex flex-wrap gap-4">
+					<LinkButton
+						href="#ecosystem"
```
bates64, line 32: Let's use the real link

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+						href="#"
+						label="Add to Edge"
+						type="secondary"
+						iconRight="ri:edge-fill"
+					/>
+				</div>
+			</div>
+		</div>
+		<div class="flex justify-center">
+			<Image src={JNLPBox} alt="JNLP-Runner logo in a tool box" />
```
bates64, line 47: Is it "JNLP-Runner" or "JNLP Runner"? The Chrome Web Store page uses the latter.
This applies to the page title too.

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,67 @@
...
+
+	<div class="grow"></div>
+
+	<!-- Scroll prompt -->
+	<a
+		href="#features"
+		class="hidden md:inline-block p-6 text-stone-400 hover:text-white transition-colors"
+	>
+		<Icon name="mdi:chevron-down" class="w-8 h-8 animate-bounce" />
+	</a>
```
bates64, line 61: We definitely need to refactor this into component and update this in the other pages. Perhaps the entire header + #after-hero can be converted into a component like

```jsx
<Hero>
    <div
		class="w-full max-w-screen-xl px-4 sm:px-8 py-16 space-y-4 md:grid grid-cols-3"
	>
		{/* ... */}
    </div>
</Hero>
```

src/components/jnlp-page/utilities/ImageSlider.astro
```diff
@@ -0,0 +1,95 @@
+---
+
+---
+
+<script defer src="https://unpkg.com/alpinejs@3.2.3/dist/cdn.min.js"></script>
```
bates64, line 5: Whoa, Alpine.js!? This is going to need some strong justification.

It looks like all of this could be done without a new dependency and with server-side rendering e.g. using Svelte's transitions. Happy to help if needed.

If we do go with it, please `pnpm install alpinejs` and import it instead of using a CDN.

src/lib/products.ts
```diff
@@ -82,6 +83,18 @@ export const products: { [product in Product]: ProductData } = {
 		github: "https://github.com/leaningtech/cheerpx-flash",
 		repositoryName: "cheerpx-flash",
 	},
+	"cheerpj-jnlprunner": {
```
bates64, line 86: TypeScript error here

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+					>
+						No Java installations
+					</h3>
+					<p class="text-base font-medium text-body-color">
+						CheerpJ JNLP-Runner is a browser extension that saves you from the
+						need of having a desktop Java installation.
+					</p>
+				</div>
+			</div>
+		</div>
```
bates64: Make this into an actual component suite, e.g. `@/components/card/Card.astro`, `@/components/card/Icon.astro`, `@/components/card/Text.astro` or something.

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+>
+	<Hero_JNLP />
+	<div
+		class="px-4 sm:px-8 py-16 space-y-4 mx-4 flex flex-wrap items-center justify-center p-10"
+		id="features"
+	>
+		<div class="px-10 md:w-1/2 lg:w-1/3">
+			<!-- Card -->
+			<div
+				class="mb-9 rounded-xl py-8 px-7 shadow-md transition-all hover:shadow-lg sm:p-9 lg:px-6 xl:px-9"
```
bates64, line 29: I don't like having a hover effect when clicking it doesn't do anything, it's weird UX.

Additionally I don't like having a shadow when the background colour of this is the same as the page background. Perhaps try a border or a differently-coloured background?

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+import layersIcon from "../../components/jnlp-page/assets/layers.png";
+import rocketIcon from "../../components/jnlp-page/assets/rocket.png";
+import screenshots from "../../components/jnlp-page/assets/screenshots.png";
+---
+
+<Shell
+	title="CheerpJ JNLP-Runner"
+	description="The browser extension for JNLP applications"
+>
+	<Hero_JNLP />
```
bates64, line 21: Underscore is horrible

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+				class="mb-9 rounded-xl py-8 px-7 shadow-md transition-all hover:shadow-lg sm:p-9 lg:px-6 xl:px-9"
+			>
+				<!-- Icon -->
+				<div class="mx-auto mb-7 inline-block">
+					<Image src={sparkleIcon} width="53" height="61" alt="sparkle icon" />
+				</div>
+				<!-- Text -->
+				<div>
+					<h3
+						class="mb-4 text-xl font-bold text-white sm:text-2xl lg:text-xl xl:text-2xl"
```
bates64, line 38: Perhaps using a slightly darker colour would be nice, the white text is a bit harsh imo

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+					iconRight="mdi:github"
+				/>
+				<LinkButton
+					href={DISCORD_URL}
+					label="Join the community"
+					type="discord"
+					iconRight="ic:baseline-discord"
+				/>
+			</div>
+		</div>
```
bates64, line 184: Expanding on what Alessandro said, the whole copy here needs rethinking I think

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+		<div></div>
+		<div class="flex justify-center items-center">
+			<Image class="scale-75" src={screenshots} alt="screenshots" />
```
bates64, line 187: It'd be really nice to incorporate screenshots of the extension into the hero. I think how I did it with my side project Mamar is a nice way of doing it: see https://mamar.nanaian.town/

src/lib/products.ts
```diff
@@ -82,6 +83,18 @@ export const products: { [product in Product]: ProductData } = {
...
+		id: "cheerpj-jnlprunner",
+		name: "CheerpJ JNLP Runner",
+		href: "/cheerpj-jnlprunner",
+		logotype: jnlprunnerLogotype,
+		favicon: "/cheerpj/favicon.png",
```
bates64, line 91: This favicon doesn't seem to be working, possibly? In any case we should definitely be using the JNLP runner's icon instead of the standard CheerpJ one!

src/lib/products.ts
```diff
@@ -82,6 +83,18 @@ export const products: { [product in Product]: ProductData } = {
 		github: "https://github.com/leaningtech/cheerpx-flash",
 		repositoryName: "cheerpx-flash",
 	},
+	"cheerpj-jnlprunner": {
+		id: "cheerpj-jnlprunner",
+		name: "CheerpJ JNLP Runner",
+		href: "/cheerpj-jnlprunner",
```
bates64, line 89: I prefer `/cheerpj-jnlp-runner`

bates64 (changes requested):
This was a really pleasant surprise, I wasn't expecting a unique page design. It's a great start.

We should think about doing something like this for the homepages of other products in the near future too as long as we can keep the docs easy to access/navigate.

src/lib/products.ts
GabrielaReyna: Since I want to have an "extensions" section on the main page, we probably need to remove the jnlp-runner from product.ts. Should we have an extension.ts file instead?

GabrielaReyna: I have added the extensions on the product.ts, added the option "type", I think would be an idea to also move the "other tools" to product.ts, what do you think?

src/pages/cheerpj-jnlprunner/index.astro
GabrielaReyna: Added :)

src/pages/cheerpj-jnlprunner/index.astro
GabrielaReyna: I came with an idea, please let me know your thoughts :)

src/components/jnlp-page/Hero-JNLP.astro
GabrielaReyna: I tried implementing this and I found it tricky since window.navigator is not accessible from Astro,  only using <script> plus I need to find a way to successfully share variables from one side to the other. I am sure there is a way but I need to dig more.
I did a workaround for now and made another type of button that looks good too. I still want to implement this idea though.

src/lib/products.ts
bates64: Agree with moving "other tools" to products.ts. Possible issue is that the site will then expect docs content for them though.

src/components/jnlp-page/Hero-JNLP.astro
bates64: Ah yeah this is a bit tricky. Perhaps we make this a svelte component and `client:idle` it. We'll need to solve the function colouring issue of Svelte not being able to use Astro components though...

bates64:
The "Latest news" section of the homepage has no padding on max width after the recent changes.

bates64:
image

I think an issue with this is that nowhere above the fold does it say "CheerpJ JNLP Runner".

bates64:
image

Not a huge fan of the dashed border, or the white background

src/components/jnlp-page/Hero-JNLP.astro
bates64: OK, let's call this out-of-scope for this PR for complexity reasons. The current solution is passable. How about we make an Asana task out of this and tackle it later?

src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -0,0 +1,70 @@
...
+	<div class="p-16 w-full relative overflow-x-hidden"></div>
+
+	<div class="grow"></div>
+
+	<!-- Scroll prompt -->
+	<a
+		href="#features"
+		class="hidden md:inline-block p-6 text-stone-400 hover:text-white transition-colors"
+	>
+		<Icon name="mdi:chevron-down" class="w-8 h-8 animate-bounce" />
```
bates64, line 66: This should be `absolute bottom-0` or something I think. Right now the scroll prompt appears below the fold, which means you'll see it _after you've already started scrolling_ which defeats the whole point!

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+import networkIcon from "../../components/jnlp-page/assets/network.png";
+import lockIcon from "../../components/jnlp-page/assets/lock.png";
+import billIcon from "../../components/jnlp-page/assets/bill.png";
+import layersIcon from "../../components/jnlp-page/assets/layers.png";
+import rocketIcon from "../../components/jnlp-page/assets/rocket.png";
+import screenshots from "../../components/jnlp-page/assets/screenshots.png";
+---
+
+<Shell
+	title="CheerpJ JNLP-Runner"
```
bates64, line 18:
```suggestion
	title="CheerpJ JNLP Runner"
```

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+	description="The browser extension for JNLP applications"
```
bates64, line 19: Could be better

bates64 (changes requested):

src/pages/cheerpj-jnlprunner/index.astro
```diff
@@ -0,0 +1,194 @@
...
+				</div>
+				<div>
+					<h3
+						class="mb-4 text-xl font-bold text-stone-100 sm:text-2xl lg:text-xl xl:text-2xl"
+					>
+						Cost-effective
+					</h3>
+					<p class="text-base font-medium text-body-color">
+						The JNLP-Runner extensions is available for a very affordable price.
+						It is an incredible option for both companies and individual users.
```
bates64, line 107: What price!?

PR #71, opened 2024-03-04
Add clientlib migration guide

DutChen18:


src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
+---
```
bates64, line 1: I think `guides` is the wrong place for this, uh, guide, at least the root of `guides` is anyway. We should probably include it as part of an upgrade guide section ([like Next.js's](https://nextjs.org/docs/app/building-your-application/upgrading)). CheerpJ sort of has this on the [Migration fro CheerpJ 2](https://labs.leaningtech.com/cheerpj3/migrating-from-cheerpj2) page. What do you think?

On a similar note, when is new clientlib shipping? Is it available on nightly today? If so, we should say so. We should be thinking about a path to stability (i.e. Cheerp 3.1) sooner rather than later too IMO although I know that's contraversial.

I'm kinda brainstorming a `version` property that you can specify in page frontmatter that defines a semver range for versions where a certain feature is supported. Then we can show a banner near the top of the page explaining the situation, or do something more clever in future.

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+title: Clientlib migration guide
+---
+
+Migrating your code can seem like a daunting task. This page aims to list
+common errors you might encounter, and contains a short description of the
+changes and features you should be aware of to make the migration easy and
+painless. This is not a complete list, but it should provide plenty of
+information to get you started, and hopefully provide some guidance if you get
+stuck.
```
bates64, line 10: Not sure how useful this paragraph really is. "Migrating your code can seem like a daunting task" is useless. I don't like describing what pages "aim" to do, that should be self-explanatory based on the table of contents on the side and the title/subtitle of the page!

More constructively:
- Perhaps write "list common errors you might encounter, and contains a short description of the changes and features you should be aware of to make the migration easy and painless" in a more succinct manner and then move it to the `description` of the page?
- "common errors you might encounter" can be a level 2 heading that wraps the common error sections (which should be level 3 headings) on this page. Perhaps "Troubleshooting"?

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+```
+
+To fix this, simply cast the union type using `->cast<T>()`. Depending on where
+the result is being used, it may be possible to omit the type parameter to the
+`cast` method.
+
+For example, for the above error, we could instead write:
+
+```cpp
+    client::ArrayBuffer* buffer = array->get_buffer()->cast();
```
bates64, line 135: Indentation

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+that is more specific than what the function returns, you can use `->cast<T>()`
+or `->cast()`.
+
+```cpp
+    client::Object* object = client::JSON.parse(string);
+```
+
+## Redefinition of client class
+
+The old clientlib was missing some modern javascript types. To use these types,
```
bates64, line 195: See Vale's comments (particularly "javascript"). We should decide on a proper casing for clientlib. Is it "clientlib" or "Clientlib", or what?

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+
+To fix the error, change the type of the argument to match the overload that
+you intend to call.
+
+```cpp
+    new client::Date(0.0); // calls double constructor
+    new client::Date(nullptr); // calls VarDate* constructor
+```
+
+## What is `_Any`, `_Union`, and `_Function`?
```
bates64, line 260: This should be on a reference page, not a migration guide

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+`EventListener*` when passing functions to higher order functions. `_Function`
+types are passed by const reference and can be constructed on the fly from raw
+function pointers and C++11 lambdas. `cheerp::Callback()` can still be used to
+get an `EventListener*`, but prefer constructing a `_Function` instead.
+
+`_Union` and `_Function` support many conversions to help keep your code clean.
+For example, `_Union<double>` can be converted to `_Union<double, String*>`,
+and `_Function<void(Object*)>` can be converted to `_Function<void(String*)>`.
+
+## What is `->cast<T>()` and `->cast()`?
```
bates64, line 290: Ditto

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+
+During migration, it can be useful to temporarily switch back to the old
+clientlib. This can be done by adding `-DUSE_OLD_CLIENTLIB` to your compiler
+flags. This flag only exists to ease the transition to the new clientlib. Do
+**not** rely on this flag to build your project. It will be removed in a future
+release.
+
+If you have some code that cannot be migrated, or encounter any other issues
+with the new clientlib, do not hesitate to reach out to us on
+[Discord](https://discord.leaningtech.com/).
```
bates64, line 327: Can use the MDX discord linkbutton here

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+
+```cpp
+    // auto cannot be inferred
+    auto* string = object->cast<client::String*>();
+
+    // type can be inferred.
+    client::String* string = object->cast();
+```
+
+## Unable to solve a migration issue?
```
bates64, line 317: I don't like having this heading be a question when the other headings are not. Perhaps `Ask for help`

src/content/docs/cheerp/guides/clientlib.md
```diff
@@ -0,0 +1,327 @@
...
+
+During migration, it can be useful to temporarily switch back to the old
```
bates64, line 319: Might be good to put this in its own little section, e.g. with the heading `Easing migration with -DUSE_OLD_CLIENTLIB`

bates64 (changes requested):
Good stuff overall. I appreciate the guide

src/content/docs/cheerp/guides/clientlib.md
DutChen18: I've moved it out of the guides section for now. I don't like having an upgrade section with only one page.

New clientlib will be on nightly Soon™.

src/content/docs/cheerp/guides/clientlib.md
DutChen18: I started writing a reference page for these features but it was getting very long and I wasn't very happy with it. The plan is to ship the new clientlib ASAP and I think just these sections in the migration guide will be good enough for now. I'll probably be coming back to this to write more extensive documentation in the future, at which point the sections in the migration guide can be replaced with a link to the actual documentation.

alexp-sssup:
The new clientlib is now merged

PR #72, opened 2024-03-08
Add TypeScript integration blog post

DutChen18:
Draft PR. I've mainly just been writing down whatever comes to mind, but the post is getting very long, so I'll likely be going over everything again at some point and cutting out a bunch of stuff. I also wrote a bit about the implementation "internals", I don't know if the is interesting enough to keep in. Still need to expand on the Clientlib section and add some introduction for the Three.js demo. Also need to figure out how to properly size the canvas for it.


src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+type __Options = { buffer: ArrayBuffer } | { absPath: String | URL };
+export default function(options?: __Options): Promise<{
+        Vec2: {
+                new(x: number, y: number): Vec2;
+        };
+        keys(_object: object): Array<string>;
+}>;
+```
+
+### Internals
```
bates64, line 123: I do like the existence of this section, although maybe it'd be better off as an expandable section, an "explanation" page on the docs, or a separate blog post? Then we link to it. Alternatively we move it to the end of the post if possible

Some readers will like it, but it's also long enough to turn others off.

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+cat library.js main.js | node
+```
+
+### Automation
+
+Writing these declarations by hand has the same issues as before. It's tedious.
+We have to write down each declaration twice. And if we ever change the
+signature of `square`, we need to update the C++ signature as well.
+
+To solve this, we have developed
```
bates64, line 190: You've changed to "we" here instead of "I". Pick one (I prefer "we")

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+---
+
+import ThreeDemo from "@/components/ThreeDemo.svelte"; 
+
+Over the past months, I have been working to add
+[TypeScript](https://www.typescriptlang.org/) integration to Cheerp. TypeScript
+is a popular programming language that adds *static type checking* to
+JavaScript.
+
+There is lots to see, so let's jump right in!
```
bates64, line 15: A TLDR would be good. Perhaps some bullet points that act as a table of contents, like
```
- Convert C++ into TypeScript declaration files
- Convert TypeScript declaration files into C++ headers
```

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+argument, but it was typed as `_Any`. This means we can also pass, for example,
+a string.
+
+```cpp
+client::square(new client::String("hello")) // returns "NaN"
+```
+
+We could manually edit the generated declaration file...
+
+```ts title="library.d.ts" ins={2} del={1}
```
bates64, line 257: V nice

bates64 (changes requested):
Loving this so far. It reads very human. 🤖 
Ofc unfinished, but I like where its going

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+declare function square(x: any): number;
+```
+
+Not bad! While it wasn't able to figure out the type of the function argument,
+it still got the correct return type.
+
+Now we can use ts2cpp to generate a C++ header.
+
+```sh
+ts2cpp library.d.ts -o library.h --pretty
```
bates64, line 215: Is ts2cpp available on npm? If not, can we make it available and use npx? Or npm install it

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
+---
+title: "TypeScript integration in Cheerp"
+description: "A tour of the TypeScript integration features that are a part of Cheerp. Cheerp is a powerful C++ to JavaScript and WebAssembly compiler."
+pubDate: "2024-03-05"
+authors: ["chen"]
+---
+
+import ThreeDemo from "@/components/ThreeDemo.svelte"; 
+
+Over the past months, I have been working to add
```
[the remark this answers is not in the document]
DutChen18: I tried writing a bit of a longer intro, but wasn't fully satisfied with what I came up with. I don't like the idea of including a commit hash in the intro either. I did rephrase it a but though.

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+Clientlib, and also discuss some of the problems we had to solve to make it all
+work.
+
+...
+
+## THREE.js demo
+
+...
+
+<ThreeDemo client:only="svelte" />
```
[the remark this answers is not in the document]
DutChen18: I've added a link to the demo in the intro.

src/content/blog/cheerp-typescript.mdx
DutChen18: After reading the entire post again, I think for some readers the internals section will actually be the most interesting part, and I think it's good to leave it in as is.

src/content/blog/cheerp-typescript.mdx
DutChen18: Added to my TODO list

src/content/blog/cheerp-typescript.mdx
DutChen18: I think the index on the right serves this purpose, no reason to duplicate the index in text.

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+    -cheerp-make-dts
+```
+
+After running this command, you should find a `square.d.ts` with content
+similar to what we wrote before.
+
+### Complex types
+
+The TypeScript declaration generator is also able to translate your own
+`[[cheerp::jsexport]]` types and types from Clientlib.
```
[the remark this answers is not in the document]
DutChen18: I've decided it's "Clientlib" and made sure to use the same capitalization everywhere.

src/content/blog/cheerp-typescript.mdx
```diff
@@ -0,0 +1,393 @@
...
+/opt/cheerp/bin/clang++ square.cpp \
+    -o square.js \
+    -target cheerp \
+    -cheerp-make-module=es6 \
+    -cheerp-make-dts
+```
+
+After running this command, you should find a `square.d.ts` with content
+similar to what we wrote before.
+
```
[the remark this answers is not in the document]
DutChen18: Added a note about IDEs, don't think a screenshot is needed.

PR #79, opened 2024-04-11
add fullWidthLayout option to docs articles

bates64:
Avoids JNLP runner page from needing to be repeated


src/components/jnlp-page/Hero-JNLP.astro
```diff
@@ -24,29 +24,29 @@ import screenshots from "@/components/jnlp-page/assets/screenshots.png";
...
+						href="/cheerpj-jnlp-runner/getting-started/install-edge"
+						label="Install for Edge"
 						type="primary2"
 						iconRight="ri:edge-fill"
 					/>
 					<LinkButton
-						href="/cheerpj-jnlprunner"
-						label="Read the docs"
+						href="/cheerpj-jnlp-runner/getting-started"
+						label="Documentation"
```
GabrielaReyna, line 49: I see what you did here! Very clever! It didn't occurred to me, I like this overlap of product page with the docs overview, avoids information from being redundant. 

Just one thing, in my screen size (Macbook pro 13") the slider button and the "Read the docs" button are very close. When I'm at the top of the page and try to click on "read the docs" it clicks on the slider instead. It may be good to space out one from the other to avoid this.

GabrielaReyna (changes requested):
I see what you did here! Very clever! It didn't occurred to me, I like this overlap of product page with the docs overview, avoids information from being redundant.

PR #82, opened 2024-05-22
Cheerp docs rewrite

bates64:
This PR improves the Cheerp documentation structure, and updates it for the defaults of Cheerp 3.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207407672218797
  - https://app.asana.com/0/0/1207420907342712


src/content/docs/cheerp/11-guides/80-clientlib-migration.mdx
```diff
@@ -1,10 +1,13 @@
 ---
 title: Clientlib migration guide
+description: How to migrate from `clientlib.h` to `client.h`
```
DutChen18, line 3: The description doesn't make sense. This suggests that `clientlib.h` was renamed to `client.h`, but both headers still exist and have the same purpose as they did before.

src/content/docs/cheerp/12-reference/05-sections/01-genericjs/index.md
```diff
@@ -0,0 +1,17 @@
...
+To place functions in the genericjs section, you can:
+
+- Tag individual functions with the `[[cheerp::genericjs]]` attribute.
+- Use `-target cheerp` to default to genericjs for all functions.
+
+### Restrictions
+
+- Cannot do type unsafe things
+  - Converting a pointer to an integer then using it is undefined behaviour
+  - Cannot convert pointers between eachother. For example, you can't read `int*` as `float*` - this will give garbage. Instead, use a union of an int and a float.
```
DutChen18, line 17: Maybe I'm being overly pedantic, but reading from an `int*` cast to `float*` is undefined in standard C++, and the value of a pointer cast to integer (even `uintptr_t`) is implementation-defined. Saying you cannot do these in genericjs doesn't make much sense to me, because you can't (or at least, shouldn't) do these things in "normal" C++ either.

The standard way of doing type punning is using `memcpy`, or (in C++20) `std::bit_cast`, of course these functions do not work in genericjs.

Maybe this section could use an example of a standard conforming C++ program that does not behave as expected using the genericjs target, and explain how and why the result is different.

src/content/docs/cheerp/12-reference/10-command-line-options/70-linear-heap-size.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "`-cheerp-linear-heap-size`"
+---
+
+`RuntimeError: memory access out of bounds` can occur if there is not enough heap memory. To increase the heap size, set `-cheerp-linear-heap-size=<value>` to a larger value (in MB, default is 1).
```
DutChen18, line 5: The default value for `-cheerp-linear-heap-size` is **8**. Maybe you got this confused with `-cheerp-linear-stack-size`, which has a default value of 1.

bates64: [Looks like this was wrong before](https://labs.leaningtech.com/cheerp/guides/interfacing-with-browser/WebAssembly-output#webassembly-heap-size) - will fix, thanks

src/content/docs/cheerp/11-guides/10-build-systems/00-make.mdx
```diff
@@ -0,0 +1,32 @@
...
+title: Make
+description: How to use Make with Cheerp
+---
+
+import LinkButton from "@/components/LinkButton.astro";
+
+To integrate Cheerp into a project that uses [Make](https://gnu.org/software/make/), use this snippet in your Makefile:
+
+```makefile
+ifeq ($(OS),Windows_NT)
```
yuri91, line 11: This is ok for  a new project, but for an existing makefile probably it makes more sense to expect to have to override CC and CXX

src/content/docs/cheerp/11-guides/10-build-systems/01-cmake.mdx
```diff
@@ -3,6 +3,8 @@ title: CMake
 description: How to build a CMake project using Cheerp
 ---
 
+import LinkButton from "@/components/LinkButton.astro";
+
 CMake projects can be compiled with Cheerp by passing a single additional command-line flag to `cmake`.
 
 A CMake toolchain file is provided at `$CHEERP/share/cmake/Modules/CheerpToolchain.cmake`, where the `CHEERP` environment variable is set to the path of your Cheerp installation.
```
yuri91, line 10: need to mention CheerpToolchain vs CheerpWasmToolchain

src/content/docs/cheerp/12-reference/10-command-line-options/10-targets.md
```diff
@@ -1,10 +1,25 @@
...
+
+## WebAssembly + JavaScript
+
+The default target, `-target cheerp-wasm`, produces both a JavaScript file and a WebAssembly file.
+
+The JavaScript file includes all C++ code compiled to JavaScript (code with the `[[cheerp::genericjs]]` attribute), and the code necessary to load the WebAssembly module and run the program.
+
+You can specify a different name and runtime path for the wasm file with the `-cheerp-secondary-output-file=<file>` and `-cheerp-secondary-output-path=<path>` options respectively.
+
+For debugging purposes, human-readable WebAssembly output can be generated using `-cheerp-linear-output=wasm`.
```
yuri91, line 20: Did you mean "human-readable asm.js output can be generated using `-cheerp-linear-output=asmjs`" ?

src/content/docs/cheerp/12-reference/40-browser-apis/index.md
```diff
@@ -0,0 +1,65 @@
...
+Browser APIs are provided by `clientlib.h`, which can be included with:
+
+```cpp
+#include <cheerp/clientlib.h>
+```
+
+## `cheerp/client.h`
+
+> [!warning] Cheerp nightly only
+> This header is currently only available on Cheerp nightly. Use `clientlib.h` on stable.
```
yuri91, line 20: this is not true. client.h was always there.
The difference with clientlib.h is that clientlib.h is JUST the generated browser APIs, while client.h also contains helpers like the SAFE_INLINE macro, the ``cheerp::*` helpers like `Callback` and makeArrayRef`, etc...

yuri91 (dismissed):
Sorry it took me a while to look at this.
Overall pretty good! Just a few minor corrections.

src/content/docs/cheerp/12-reference/10-command-line-options/10-targets.md
bates64: This text was there before. I can change this to what you suggest if you prefer. Does `-cheerp-linear-output=wasm` not exist, then?

yuri91: it does exist, but it's the default

yuri91 (approved):

PR #88, opened 2024-06-24
Cheerpj applet runner page

Honggeiry:
Simple applet runner page implementation

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207421663310575


GabrielaReyna (changes requested):
There are multiple .history files on the directories you have been modifying. I believe this shouldn't happen. We can take a look together on why it's happening.

GabrielaReyna (changes requested):
Hiya, about the prettier error:

Prettier check can be solved by formatting the code of the problematic files. 
You can do it with the command `pnpm format` to format all of them. Or you can format individual files or sections by selecting the codelines/file and right click > format with > Prettier or `CMD + Shift + P`.

You can also locally check that formatting is correct with the command: `pnpm prettier . --check`.

PR #100, opened 2024-07-27
cheerp home 

GabrielaReyna:
Worked on first iteration, needs more refining and sections


packages/astro-theme/components/LinkButton.astro
```diff
@@ -3,11 +3,10 @@ import { Icon } from "astro-icon";
 interface Props {
 	href: string;
 	label: string;
-	type?: "primary" | "primary2" | "secondary" | "discord";
+	type?: "primary" | "primary2" | "secondary" | "primary-cheerp" |"secondary-cheerp"| "discord";
```
bates64, line 6: You could probably do away with this by changing the primary colour in the Tailwind config to #8e88ff for Cheerp. So all buttons and links change to the Cheerp theme, not just specific buttons

sites/cheerp/src/pages/index.astro
```diff
@@ -1,37 +1,45 @@
...
 
-<ProductHome>
-	<section class="py-16">
+<ProductHome style = "px-6 py-10 w-full  mx-auto bg-gradient-to-b to-[#2f8f8a] from-stone-900">
+
+	<section class="px-48 flex col-span-2 items-center">
 		<div class="mx-auto max-w-[56em] text-center space-y-12">
 			<h1
-				class="text-7xl font-semibold text-balance bg-gradient-to-b from-cheerp via-cheerp to-white text-transparent bg-clip-text"
+				class="text-white text-4xl md:text-5xl lg:text-6xl xl:text-7xl font-semibold my-4 text-left"
```
bates64, line 13: Add text-balance to make the text wrap more cleanly.

text-wrap: initial
image

text-wrap: balance
image

packages/astro-theme/layouts/ProductHome.astro
```diff
@@ -6,6 +6,7 @@ import BentoGrid from "../components/patterns/BentoGrid.astro";
 import { productFromUrl } from "../lib/products";
 
 const product = productFromUrl(Astro.url)!;
+const {style = "px-6 py-10 w-full  mx-auto bg-gradient-to-b to-[#ffffff44] from-stone-900"}= Astro.props;
```
bates64, line 9: I'd probably make all of these classes default and merge in a `class` prop using clsx

sites/cheerp/src/pages/index.astro
```diff
@@ -1,37 +1,45 @@
...
 import LinkButton from "@leaningtech/astro-theme/components/LinkButton.astro";
+import heroAsset from "../assets/cheerpcube.png";
+import { Image } from "astro:assets";
 ---
 
-<ProductHome>
-	<section class="py-16">
+<ProductHome style = "px-6 py-10 w-full  mx-auto bg-gradient-to-b to-[#2f8f8a] from-stone-900">
+
+	<section class="px-48 flex col-span-2 items-center">
```
bates64, line 10: This is not responsive

bates64 (changes requested):

packages/astro-theme/components/LinkButton.astro
GabrielaReyna: I'm still unsure if I will use the color somewhere else that is not the home site, will have this in mind if I keep on using the color somewhere else as I progress with the design. Good idea.

sites/cheerp/src/pages/index.astro
```diff
@@ -1,61 +1,243 @@
...
+			</div>
+			<div class="col-span-1 border-solid border-l text-left py-4">
+				<p class="text-left text-2xl font-bold text-stone-300 pl-8">
+					<span class="text-white">Ready to build?</span> Get started today by browsing
+					our
+					<a href="/docs" class="text-cheerp hover:text-white">documentation</a>
+					and
+					<a href="/docs/tutorials" class="text-cheerp hover:text-white"
+						>tutorials</a
+					>
```
Honggeiry, line 224: I noticed that the documentation and tutorials are jump links, but they are not very obvious, because the ‘Cheerp Enterprise’ text on the left is the same style but not a link, so I thought maybe it might be better to put external link indicators next to them, here is a solution I found for your reference:

```suggestion
					<a
						href="/docs"
						class="text-cheerp hover:text-white inline-flex items-center"
						target="_blank"
						>documentation
						<svg
							class="w-4 h-4 ml-1"
							fill="currentColor"
							viewBox="0 0 20 20"
							xmlns="http://www.w3.org/2000/svg"
						>
							<path
								d="M11 3a1 1 0 100 2h2.586l-6.293 6.293a1 1 0 101.414 1.414L15 6.414V9a1 1 0 102 0V4a1 1 0 00-1-1h-5z"
							></path>
							<path
								d="M5 5a2 2 0 00-2 2v8a2 2 0 002 2h8a2 2 0 002-2v-3a1 1 0 10-2 0v3H5V7h3a1 1 0 000-2H5z"
							></path>
						</svg>
					</a>
```

Honggeiry (changes requested):
Great job on the home page! I really like the colour theme and that open sourcerers part, very cool! 

I left two comments, the button width uniformity one I think is necessary, and the one suggesting external links is just a suggestion. Anyway it's really great! Thanks for the hard work!

Honggeiry (approved):
Thank you for the great work!

PR #104, opened 2024-07-31
fix li margin on cheerpj.com navbar

bates64:
The WordPress theme on cheerpj.com adds a bottom margin to all `li` elements by default, which we don't want

image


bates64:
Closing in favour of #105. Probably delete this branch

PR #105, opened 2024-07-31
global-navbar fixes

bates64:
1. Fixes the WordPress theme on cheerpj.com adding a bottom margin to all `li` elements:
image
2. Fixes lack of styling on labs due to the way Svelte postprocesses CSS.
3. Makes it possible to select Cheerp on the mobile nav


Honggeiry (approved):
LGTM

PR #108, opened 2024-08-05
Update Cheerp web workers tutorial

DutChen18:
This tutorial did not work because, with the code as it was written before, a message was being sent to a worker before it was able to register a handler for it, causing the message to be lost. I've simplified the example by not having the worker wait for a message, instead it immediately posts its own messages to the main thread.

I've also updated the example to work with the new clientlib.


GabrielaReyna (changes requested):
Hello!  Could you please change all the h1 to h2 and so on so:
 - `# title ` becomes ` ## title`
 - `## title2 ` becomes `### title2 `and so on. 
 
Thanks for the PR!

GabrielaReyna (approved):

PR #109, opened 2024-08-05
fixing images on docs

Honggeiry:


GabrielaReyna (changes requested):
I think the paths that contain `labs/public` specially in the CheerpJ docs won't work. They will work locally because you have the full repo in your localhost. As far as I remember CheerpJ docs are not served from the same place as the rest of the docs.

For instance, if the basepath for cheerpj docs start with /docs then the images in public for cheerpj should be accessed something like "/docs/cheerpj3/assets/..." which is very strange and I'm not 100% it that will work but we need to test it. 

Let's talk about it tomorrow.

GabrielaReyna (changes requested):
Thanks for fixing the paths for CheerpX, I think I found what's the issue for the specific case of the extensions docs:
There is a redirect of anything that starts with `https://labs.leaningtech.com/cheerpj3/` to become `cheerpj.com/docs` (because that's the old url before the monorepo and there are bunch of links out there that still uses the older links).

Given the assets in public for the JNLP and Applet runner extensions are under:

- `sites/labs/public/cheerpj3/assets/jnlp-screenshots`

After the labs site is built, everything under public is copied to the root of the site, meaning these images are found as:
- `/cheerpj3/assets/jnlp-screenshots`
When we are trying to serve these images, these are served as:
- `labs.leaningtech.com/cheerp3/assets/jnlp-screenshots` (because the root of the site becomes `labs.leaningtech.com`)

Since there is redirect rule, then this images are actually trying the be found in the wrong place because the url is converted to `cheerpj.com.com/docs/assets/jnlp-screenshots`

So the fix should be to remove the named `cheerpj3` and place the content one level above so the assets for the extensions are:
- `sites/labs/public/assets/applet-screenshots`
- `sites/labs/public/assets/jnlp-screenshots`

And then update the paths to the correct ones on the markdown files. For example:
- `/assets/applet-screenshots/image_name.png`

I hope is clear, let me know if you need anything :)

GabrielaReyna (approved):

PR #115, opened 2024-08-26
Cheerpx.io home page

GabrielaReyna:


sites/cheerpx/src/pages/index.astro
```diff
@@ -1,60 +1,183 @@
...
+					label="Contact us"
+				/>
+			</div>
+			<div class="col-span-1 border-solid border-l text-left py-4">
+				<p class="text-left text-2xl font-bold text-stone-300 pl-8">
+					<span class="text-white">Ready to build?</span> Get started today by browsing
+					our
+					<a
+						href="/docs"
+						class="text-yellow-600 hover:text-white inline-flex items-center"
```
Honggeiry, line 148: `class="text-yellow-600` You may have forgotten to change the font colour here, it's still in orange, you can pick a font colour you feel appropriate here: https://tailwindcss.com/docs/text-color

sites/cheerpx/src/pages/index.astro
```diff
@@ -1,60 +1,183 @@
...
-			/>
-			<LinkButton
-				type="secondary"
-				href="https://webvm.io"
-				iconRight="mi:arrow-right"
-				label="Try the demo"
-			/>
+		<div class="flex justify-center items-center">
+			<LinkButton type="primary-cheerpx" href="#youtube" label="Learn more" />
 		</div>
```
Honggeiry, line 91: Below are just some advice:

Because the text above says ‘For now, enjoy the beta trial and our popular WebVM demo.’ I think it would be good to add another button like ‘try our WebVM demo’ here next to the learn more button to attract people to click on it.

Also, we can edit the text to simply describe the Games Runner Extension and add another ‘Try our Games Runner Extension’ button here as well.

(Or, If there's enough bandwidth, we can also definitely list WebVM and Games Runner Extension in a separate section)

Honggeiry (changes requested):
Hi Gaby thanks for your great work on this! 

I like it a lot, but there is one place where you may need to change the font colour. The other one is just out of my personal suggestion. Also I totally agree with ‘one more section for the hackathon when the time comes ’.

sites/cheerpx/src/pages/index.astro
```diff
@@ -1,60 +1,243 @@
...
+					class="w-full max-w-[1080px] aspect-video rounded-xl shadow-lg border-2 border-stone-900"
+					frameborder="0"
+					allowfullscreen="allowfullscreen"
+				>
+				</iframe>
+			</div>
+			<div class="col-span-1 pl-8">
+				<h1
+					class="text-white text-5xl xl:text-5xl 2xl:text-6xl font-semibold py-4 text-left text-wrap md:text-balance"
+				>
```
Honggeiry, line 170: We might need to get the text here to be centralised rather than left-aligned on the phone screen to show it fully. Or alternatively we can reduce the size of the h1 font here

Honggeiry (commented):
Thank you for making the changes! The newly added CheerpX Games Runner section looks great! The comments I left are all opinions, feel free to use them or not as you see fit :)

sites/cheerpx/src/assets/HDgamepad.png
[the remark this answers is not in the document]
GabrielaReyna: I think the whole logo would be an excessive use of the shield. It is already in the top-left corner of the page and on the hero design.

sites/cheerpx/src/assets/screenshots.png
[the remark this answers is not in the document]
GabrielaReyna: This makes sense and it was my first approach but it didn't look good given the background colour of the extension page didn't match the theme on the webpage. I'll leave like it is for now.

sites/cheerpx/src/pages/index.astro
GabrielaReyna: Good idea, thanks. This whole section will be different, I'll be removing the youtube video, probably I will leave the title in a slightly different spread. Thanks for the review!

Honggeiry (approved):
Thank you and great job on these changes!

PR #123, opened 2024-09-10
Updated cheerpx docs, added licensing page

Honggeiry:


sites/cheerpx/src/content/docs/12-reference/00-CheerpX-Linux-create.md
```diff
@@ -0,0 +1,163 @@
...
+
+```js
+const idbDevice = await CheerpX.IDBDevice.create("dbName");
+
+const cx = await CheerpX.Linux.create({
+	mounts: [{ type: "tree", path: "/files", dev: idbDevice }],
+});
+```
+
+This setup creates a virtual filesystem at `/files` that is backed by IndexedDB.
```
saarikoski-jules, line 148: For consistency, I think you could also mention WebDevice and DataDevice initialization here (or split the devices up into a different page? Not sure what makes the most sense). CheerpX.DataDevice.create() takes no params, CheerpX.WebDevice.create("") takes a path to a directory in the local server (Technically external urls should also work but this hasn't been tested, or really intended use at this point).

sites/cheerpx/src/content/docs/12-reference/00-CheerpX-Linux-create.md
```diff
@@ -0,0 +1,163 @@
...
+
+### `dataDevice.writeFile`
+
+CheerpX also provides a method to write data to files within the mounted devices. This utility is limited to creating or updating files at the root level. Note that this feature is still subject to review.
```
saarikoski-jules, line 152: writeFile cannot update files, it can only create a new, previously nonexisting file.

You should remove the subject to review portion of this. When the new API is published the feature will remain as is.

saarikoski-jules (commented):
I wasn't able to comment on the licensing portion, you should seek a review from someone else for that

sites/cheerpx/src/content/docs/12-reference/00-CheerpX-Linux-create.md
Honggeiry: I have added examples for WebDevice and DataDevice alongside IDBDevice in the same section. For now I'll just leave them on the same page since they are all related to CheerpX.Linux.create. (Of course if we don't feel right about it later, we can always change it later)

sites/cheerpx/src/content/docs/12-reference/00-CheerpX-Linux-create.md
```diff
@@ -0,0 +1,213 @@
...
+const cx = await CheerpX.Linux.create({
+	mounts: [{ type: "tree", path: "/data", dev: dataDevice }],
+});
+```
+
+This creates an in-memory filesystem mounted at `/data`.
+
+### `dataDevice.writeFile`
+
+CheerpX provides a method to write data to new files within the mounted devices. This utility is limited to creating files at the root level of the device.
```
saarikoski-jules, line 186: "CheerpX.DataDevice provides a method ...", as this is the only device that provides such a utility.

sites/cheerpx/src/content/docs/11-guides/02-custom-devices.md
GabrielaReyna, line 1: Could you please make the image on step #3 smaller?

GabrielaReyna (changes requested):
Thanks for working on this. This reads well in general and I'll trust Jules previous review for the technical aspect. 

I noticed potential improvement points that may require a full guide page but we can add that later on the go. For example, a guide on the file system mount points, just as we did for CheerpJ

More than the image resize, I consider this ready to merge and I'll double check with Stefano about the licensing part.

GabrielaReyna (approved):

PR #136, opened 2024-10-08
Cheerpx docs update review

alexp-sssup:


sites/cheerpx/src/content/docs/12-reference/00-CheerpX-Linux-create.md
```diff
@@ -67,6 +67,9 @@ const cx = await CheerpX.Linux.create({
 });
 ```
 
+> [!note]
```
GabrielaReyna, line 70:
```suggestion
> [!note] Note
```

GabrielaReyna (changes requested):

sites/cheerpx/src/content/docs/12-reference/00-CheerpX-Linux-create.md
GabrielaReyna: There is another callout on the same page without title but cannot suggest a commit there.

PR #141, opened 2024-10-09
Updates to full os tutorial

saarikoski-jules:


alexp-sssup:
All the nginx config for the image slicing and the images directly are absolutely not needed, they are leftovers from old setup. The bytes backend just works with no additional config.

Also, try to make it more clear that all the various snippets of JavaScript needs to be placed in the same script tag created for the first snippet.

PR #151, opened 2024-10-21
Changed CONTRIBUTING.md type

AtibQur:
I called the branch documentation-page instead of contributing-page..

I moved the CONTRIBUTING.md page to the front because it wasn't showing any content. If the contributing file was not needed in: sites/labs/src/content/docs/CONTRIBUTING.md. Than this should be fine now.


alexp-sssup:
> If the contributing file was not needed in: sites/labs/src/content/docs/CONTRIBUTING.md

Do we know if it was, actually, not needed. If it was in that location there must have been a reason. `grep` for the file name across the code and inspect git history with `git log` for that file and the symlink. Try to see if there is information about why the file was in that location.

AtibQur:
I’ve checked the logs and commits but I couldn't make sense out of it why it's created in that location.

I found two contribution pages: one in the labs directory and another in the cheerp directory. The labs contribution page is currently linked in the directory. I can't seem to find any other reason (next to the two pages) why it's in that location.

alexp-sssup:
I think it's used here, please investigate `packages/astro-theme/lib/nav.ts`. Search for `CONTRIBUTING`

AtibQur:
When the contributing link gets clicked in the footer component, the website redirects to the contributing page that is in labs/docs. The link is broken if the file is moved to another place.

I think the location of the contribution page is okay. I'm closing this pr.

PR #154, opened 2024-10-26
Added info to nginx doc

AtibQur:


sites/cheerpx/src/content/docs/11-guides/nginx.md
```diff
@@ -43,9 +49,16 @@ http {
 }
 ```
 
-This configuration will work over localhost in a development environment. When the application is deployed, it must be served over HTTPS.
+This is a basic nginx configuration that allows CheerpX to run.
+
+### 2. Important notes
+
+- This setup works in a **development environment** on `localhost`
+- **For production**, always use **HTTPS** to ensure security.
```
alexp-sssup, line 57: It's not fully accurate, it's actually not possible to have cross origin isolation without https. Localhost is an exception to the rule

alexp-sssup (changes requested):

alexp-sssup (approved):

PR #160, opened 2024-10-31
Prettier configuration changed

AtibQur:


package.json
```diff
@@ -10,7 +10,7 @@
 		"typescript": "^5.2.2"
 	},
 	"prettier": {
-		"trailingComma": "all",
+		"trailingComma": "none",
```
alexp-sssup, line 13: Is there a way to only remove the trailing commas for functions, but not on objects or arrays? I would like to minimize the impact on the repo

alexp-sssup (changes requested):

package.json
AtibQur: There is no option for that with prettier. You can only set trailing commas for "all", "es5" or "none".

alexp-sssup: Please see what is the impact of es5. The objective is to remove trailing commas in function calls, since it's really confusing and sloppy. If es5 does not do what we want we can use none

AtibQur: "es5"  keeps trailing commas for objects and arrays, but does not add them for function parameters.
I think "none" is the closest to what we want.

alexp-sssup: > "es5" keeps trailing commas for objects and arrays, but does not add them for function parameters.
This seems pretty much what I want

AtibQur: Ah than I misunderstood! I'll be fixing it right now

alexp-sssup (approved):

PR #162, opened 2024-11-04
Version bump

AtibQur:


alexp-sssup:
Please rework this to have a single variable controlling the version everywhere

sites/cheerpx/src/content/docs/10-getting-started/index.mdx
```diff
@@ -23,12 +28,13 @@ CheerpX is also available as an ES6 JavaScript module. If you prefer to use the
 
 ```html
 <script type="module">
-	import * as CheerpX from "https://cxrtnc.leaningtech.com/1.0.0/cx.esm.js";
+	// Insert the version number of CheerpX
+	import * as CheerpX from "https://cxrtnc.leaningtech.com/{VERSION_NUMBER}/cx.esm.js";
```
alexp-sssup, line 32: This does not work, as expected
image

Please use a local build, it does not make sense for me to review a not yet working solution

alexp-sssup (changes requested):
Current solution does not work. Try to get a better understanding about how the ``` actually works. We want to get the same rendering style using HTML directly.

AtibQur:
I think I've fixed the problem. I'm closing this PR/branch and will make a new PR with another branch

PR #167, opened 2024-11-11
updated native methods guide

theodoravraimakis:


sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Implementing Native Methods
+description: Java Native Interface (JNI) with CheerpJ
+---
+
+CheerpJ allows you to implement Java native methods (typically written in C/C++ or another AOT-compiled language) directly in JavaScript, similar to the Java Native Interface (JNI) in standard Java.
+
+In Java, native methods are identified by the `native` keyword in their declaration. These methods are not implemented in Java but are instead defined in an external language, which in the case of CheerpJ, is JavaScript.
+
```
epanholz, line 9: We could already add the steps that we will explain here, something like "Adding natives to CheerpJ only requires a few steps: 1. Define a native function in Java 2. Implement it in Java Script 3. Pass it to CheerpJ". You could also a bit of context why natives would be useful.

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,103 @@
...
+## Implementing Native Methods in JavaScript
+
+To implement a native method in JavaScript, create an `async` function that follows the naming convention `Java_<ClassName>_<MethodName>`. The function takes:
+
```
epanholz, line 13: We should add a part here that already shows how to add the native function in Java, maybe a header before the `Implementing Native Methods in JavaScript` and call it something like `How do add a native function to Java` and just show the same example class.

The old docs also had a part about the naming that was very clear, please add this:
`Java_<fully-qualified-class-name>_<method-name>` - For instance, if com.foo.Bar has a native method called `baz`, its object key is `Java_com_foo_Bar_baz`.

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,103 @@
...
+Here, we provide an implementation for the `alert` method in the `Example` class, by creating a function named `Java_Example_alert`. The function displays an alert dialog with the message using `window.alert`.
+
+3. Initialize CheerpJ with the `natives` option and pass the native method implementation to `cheerpjInit`:
+
+```js title="index.html"
+// Native method implementation
+async function Java_Example_alert(lib, str) {
+	window.alert(str);
+}
+
```
epanholz, line 93: We should add the full HTML code here to make it clear where the function is defined.

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,103 @@
...
+public class Example {
+  public static void main(String[] args) {
+    alert("Hello, world!");
+  }
+
+  public static native void alert(String message);
+}
+```
+
+2. Implement the native method by creating an `async` function in JavaScript that follows the naming convention `Java_<ClassName>_<MethodName>`.
```
epanholz, line 75: You should use `Java_<fully-qualified-class-name>_<method-name>` instead of `Java_<ClassName>_<MethodName>` to make it more clear.

epanholz (approved):
This one looks good to me know. For the future please `--amend` to the first commit if I request some changes, to avoid unnecessary commits. Keeps the history clean.

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,135 @@
+---
+title: Implementing Native Methods
+description: Java Native Interface (JNI) with CheerpJ
+---
+
+CheerpJ allows you to implement Java native methods (typically written in C/C++ or another AOT-compiled language) directly in JavaScript, similar to the Java Native Interface (JNI) in standard Java.
```
GabrielaReyna, line 6:
```suggestion
CheerpJ allows one to implement Java native methods (typically written in C/C++ or another AOT-compiled language) directly in JavaScript, similar to the Java Native Interface (JNI) in standard Java.
```

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,135 @@
...
+
+  public static native void alert(String message);
+}
+```
+
+The method is defined in the Java class but is not implemented in Java. Instead, the implementation will be provided in JavaScript.
+
+## Implementing Native Methods in JavaScript
+
+To implement a native method in JavaScript, create an `async` function that follows the naming convention `Java_<fully-qualified-class-name>_<method-name>`. For instance, if com.foo.Bar has a native method called baz, its object key is Java_com_foo_Bar_baz. The function takes:
```
GabrielaReyna, line 36:
```suggestion
To implement a native method in JavaScript, create an `async` function that follows the naming convention `Java_<fully-qualified-class-name>_<method-name>`. For instance, if `com.foo.Bar` has a native method called `baz`, its object key is `Java_com_foo_Bar_baz`. The function takes:
```

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,135 @@
...
+		</script>
+	</body>
+</html>
+```
+
+In this setup, `cheerpjInit` loads `Java_Example_alert` as the native method implementation. When `Example.alert` is called in Java, it triggers the JavaScript `Java_Example_alert` function, displaying an alert dialog with the message.
+
+[`natives`]: /docs/reference/cheerpjInit#natives
+[`CJ3Library`]: /docs/reference/CJ3Library
+[`conversion rules`]: /docs/reference/CJ3Library#conversion-rules
```
GabrielaReyna, line 135:
```suggestion
[`conversion rules`]: /docs/reference/CJ3Library#conversion-rules
[`cheerpjInit`]: /docs/reference/cheerpjInit
```

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,135 @@
...
+  // Implementation
+}
+```
+
+> [!info] Handling Static Native Methods
+> If the native method is static, the `self` parameter can be omitted.
+
+## Passing Native Functions to CheerpJ
+
+To use the native method in CheerpJ, pass the function to the `cheerpjInit` function as a property of the [`natives`] option. You can pass:
```
GabrielaReyna, line 55:
```suggestion
To use the native method in CheerpJ, pass the function to the [`cheerpjInit`] function as a property of the [`natives`] option. You can pass:
```

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,135 @@
...
+
+```js title="index.html"
+async function Java_Example_alert(lib, str) {
+	window.alert(str);
+}
+```
+
+Here, we provide an implementation for the `alert` method in the `Example` class, by creating a function named `Java_Example_alert`. The function displays an alert dialog with the message using `window.alert`.
+
+3. Initialize CheerpJ with the `natives` option and pass the native method implementation to `cheerpjInit`:
```
GabrielaReyna, line 109:
```suggestion
3. Initialize CheerpJ with the `natives` option and pass the native method implementation to [`cheerpjInit`]:
```

sites/cheerpj/src/content/docs/11-guides/implementing-native-methods.md
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			await cheerpjInit({ natives: { Java_Example_alert } });
+			await cheerpjRunMain("Example", "/app/");
+		</script>
+	</body>
+</html>
+```
+
+In this setup, `cheerpjInit` loads `Java_Example_alert` as the native method implementation. When `Example.alert` is called in Java, it triggers the JavaScript `Java_Example_alert` function, displaying an alert dialog with the message.
```
GabrielaReyna, line 131:
```suggestion
In this setup, [`cheerpjInit`] loads `Java_Example_alert` as the native method implementation. When `Example.alert` is called in Java, it triggers the JavaScript `Java_Example_alert` function, displaying an alert dialog with the message.
```

GabrielaReyna (changes requested):
Thanks for the PR and great work. I have suggested a couple ofchanges regarding formatting. 

I'd like to ask about the difference between this guide and "Implementing native methods and libraries". We will probably need to rename these guides to make it less confusing for the user since them two have very similar titles.

epanholz:
We decided to split the "Implementing native methods and libraries" guide into two separate ones, we heard from some customer that it is a bit difficult to understand so we also revamped them a bit. There is another PR with the standalone libraries guide.

GabrielaReyna (approved):

PR #169, opened 2024-11-11
added guide on implementing native libraries and removed the original

theodoravraimakis:


epanholz (changes requested):
Another round of review, after these changes this guide should be done. Please add the changes with `git commit --amend` to avoid adding too many commits.

sites/cheerpj/src/content/docs/11-guides/implementing-native-libraries.md
```diff
@@ -0,0 +1,140 @@
...
+
+CheerpJ adapts this concept to load JavaScript modules, using the same `System.loadLibrary` call to load `.js` files. These files are JavaScript modules that define implementations for the native methods associated with the library.
+
+## Overview
+
+In general, we can implement native libraries in CheerpJ by following these steps:
+
+1. Create a JavaScript module that implements the native methods.
+2. Create a Java class that loads the native library and declares the native methods.
+3. Load the native library in the Java class with CheerpJ.
```
epanholz, line 24: Sorry for the back and forth here, I think it might be more clear if we swap section 1 & 2. So it would look like this:
> 1. Create a Java class that loads the JavaScript library and declares the native methods.
> 2. Create a JavaScript module that implements the native methods.
> 3. Load the naive library in the Java class with CheerpJ.

Make sure to also swap the sections and explanations.

sites/cheerpj/src/content/docs/11-guides/implementing-native-libraries.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Creating a JavaScript Module
+
+JavaScript functions that implement native methods should follow a specific naming convention - `Java_<fully-qualified-class-name>_<method-name>`. For instance, if `com.foo.Bar` has a native method called `baz`, the function will be called `Java_com_foo_Bar_baz`
```
epanholz, line 28: I would add a small explanation on what JavaScript modules are. Maybe like this?
> A JavaScript module is a file that contains code which can be exported and imported by other files for better organization and reuse. You create modules using the `export` keyword to expose classes, methods, or other resources, and you can use `export default` to make a primary, easy-to-import item from the module.

You could then also link to the official documentation like this:
For more information on JavaScript modules, check out the official [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).

You could add a code snippet looking like this:
```
module.js
export default {
      // Java Script code
};
```
After that you can continue with this part again: JavaScript functions that will implement native Java functions need to follow a specific naming convention etc.

sites/cheerpj/src/content/docs/11-guides/implementing-native-libraries.md
```diff
@@ -0,0 +1,140 @@
...
+```js
+async function Java_<fully-qualified-class-name>_<method-name>(lib, self, param1, param2, ...) {
+  // Implementation
+}
+```
+
+> [!info] Handling Static Native Methods
+> If the native method is static, the `self` parameter can be omitted.
+
+### Native Methods in Java
```
epanholz, line 47: We should also add the library part to the header here, maybe:
### Loading native libraries and declaring native methods in Java

sites/cheerpj/src/content/docs/11-guides/implementing-native-libraries.md
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  // Native method declaration
+  private native void methodName(param1, param2, ...);
+}
+```
+
+### Initializing CheerpJ with the `javaProperties` Option
+
+To use a native library in CheerpJ, you need to set the `java.library.path` property to the directory containing the JavaScript file that implements the native methods. This is done by passing the `javaProperties` option to the `cheerpjInit` function:
```
epanholz, line 65: JavaScript file -> JavaScript module file

epanholz (changes requested):
Sorry for the back and forth here, after these changes are implemented this guide should be ready too. I really want the new docs to be clear and easy to understand. Don't forget to use `git commit --amend` to apply the changes and then force push to your branch again.

epanholz (approved):
This looks good to me now except one typo.
@GabrielaReyna please have a final look before merging.

GabrielaReyna (approved):
Thanks for the changes, this is ready to merge.

PR #176, opened 2024-11-19
Added blog/video to CheerpX

AtibQur:
With this added, the video works also when in the CheerpX blog section. The issue was that the path was not correct so it wasn't able to load any video at all.

I tried adding the right path to the .mp4 file in labs but video <source> tag is not able to reach it when loading the website from the CheerpX website. Are all the branches (Cheerp, CheerpJ, CheerpX, labs) isolated from each other?


GabrielaReyna:
I think you need to move the video to the same folder where the blog .md file is (´/content/blog/´).
Make sure that the path in the markdown file is also correct once you do it. 

Not ´sites/CheerpX/...´ and not 'sites/labs/public/blog/alpine-mines.mp4'

GabrielaReyna (approved):

PR #181, opened 2024-11-26
ignore relevancy on blog post display for labs

GabrielaReyna:


AtibQur (approved):
It appears to be working as intended. The blog is functioning correctly.

PR #194, opened 2024-12-06
Added more context to the new CJ native method docs.

epanholz:
I added a bit more context to the new tutorial and guide. I renamed the native java function from `alert()` to `nativeAlert()` to avoid confusion with JavaScript method of the same name. I also added a part to the guide that explains how to call back to Java using the `lib` parameter and referenced that in the tutorial.


GabrielaReyna (approved):
Thanks for the PR, this looks good to go :)

PR #195, opened 2024-12-07
Added processCreated section

AtibQur:


sites/cheerpx/src/content/docs/12-reference/20-CheerpX-linux-callbacks.md
```diff
@@ -85,3 +85,32 @@ if (cx) {
...
+```
+
+2. **Callback implementation**
+
+The `handleProcessCreated` function increments a `processCount` variable and triggers a callback if one is set via the processCallback function.
+
+```js
+function handleProcessCreated() {
+	processCount++;
+	if (processCallback) processCallback(processCount);
```
alexp-sssup, line 112: Calling another callback is redundant here, this is already a callback and the user can run arbitrary logic here

alexp-sssup (changes requested):

alexp-sssup (approved):

GabrielaReyna (changes requested):
Thanks for adding this information. A couple of things:
- I am finding this page hard to navigate. There may be a better way to organise subtitles in order to show the subsections on the table of contents. For example, making the parameters and returns nested subtitles. 
- I  understand the info here is dense and required it's own page. Please link this page reachable from ´/CheerpX-Linux-create#returns´. I'm also thinking if part of this should go under CheerpX-Linux-create + a guide. Let's discuss this with Alessandro.

GabrielaReyna (approved):

PR #197, opened 2024-12-10
dirname with dot

GabrielaReyna:
Please do not merge yet


GabrielaReyna:
@AtibQur This is the PR I told you about where I'm fixing the dot in a dirname. Could you please help me to test it locally? (E.g add madeup files using the structure of the reference we will be adding). You don't need to commit those changes. Just check that it doesn't break anything please. I have done some tests myself but just to be sure.

PR #199, opened 2024-12-14
Added edit /etc/subgid section

AtibQur:


sites/cheerpx/src/content/docs/11-guides/custom-images.mdx
```diff
@@ -81,3 +81,49 @@ Remove temporary directories:
...
+```
+
+- `<username>` is the name of the user,
+- `<start_id>` is the starting ID of the range of subgroups assigned to the user,
+- `<count>` is how many consecutive subgroup IDs the user is assigned.
+
+For user with PID 65534 (typically the nobody user), the entry would look something like:
+
+```
+nobody:100000:65536
```
alexp-sssup, line 110: Wait, I am not sure about this. I have a vague recollection that the user that needs to be changed is the _current_ user, the id inside the container is irrelevant here. Can you validate?

sites/cheerpx/src/content/docs/11-guides/custom-images.mdx
```diff
@@ -81,3 +81,49 @@ Remove temporary directories:
...
+```
+nobody:200000:65536
+```
+
+### 4. Save and apply changes
+
+Save the file and ensure changes are recognized by restarting your container service:
+
+```bash
+sudo systemctl restart podman
```
alexp-sssup, line 128: Is this actually required?

alexp-sssup (changes requested):

alexp-sssup:
@AtibQur Can you confirm this has been validated as working?

AtibQur:
@alexp-sssup for some reason I keep getting Error: cannot setup namespace using newuidmap: exit status 1. 

This happens when I'm using: podman run --rm -it --userns=keep-id alpine to start a podman container with user namespace even though uidmap is already installed on my system.
Do you have any idea/experience how to deal with this or how to test it in another way?

alexp-sssup:
I have never seen this error and I think `--userns=keep-id` is not correct, I certainly did never use it

AtibQur:
@alexp-sssup I've been testing changes to /etc/subgid while running podman containers to see how group IDs are mapped. The changes are working as expected for /etc/subgid.
I was wondering if I need to also write about changing /etc/subuid to ensure proper mapping for both UIDs and GIDs

alexp-sssup (approved):

sites/cheerpx/src/content/docs/11-guides/custom-images.mdx
```diff
@@ -81,3 +81,45 @@ Remove temporary directories:
 ```bash
 rm -rf cheerpXFS
 ```
+
+## Edit `/etc/subgid` (for the host user)
```
GabrielaReyna, line 85: I noticed this guide has a step 1,2,3 format: 

- Is this section a step within the 1,2,3 etc? If so, please add the number. If is not a mandatory step but an optional action, please specify so within the text in the first paragraph of this section. 
- Please also specify at the end of the "Last step" that until that point things should work or that the setup is complete and what the user can expect at the end of the guide.

GabrielaReyna (changes requested):

GabrielaReyna (approved):

PR #202, opened 2024-12-18
CJ docs updates

epanholz:
We only want to merge these doc changes once we release CheerpJ 3.1, so not a high priority for now.
I added documentation for some new cheerpjInit options and added a small guide on how to use one of the new options.


GabrielaReyna (changes requested):

sites/cheerpj/src/content/docs/12-reference/00-cheerpjInit.md
```diff
@@ -373,6 +375,48 @@ Example of usage:
...
+```
+
+This option enables advanced debug logging, which is helpful for troubleshooting issues with CheerpJ.
+
+Example of usage:
+
+```js
+cheerpjInit({ enableDebug: true });
+```
+
```
GabrielaReyna, line 419: Could you please add a short line here similar to "Learn more about debugging on our guide [link to the guide]" the same way you did for execCallback .

GabrielaReyna (changes requested):

PR #203, opened 2025-01-01
Fix device name in "Reading Files from JavaScript" example

hubgit:


alexp-sssup:
Thanks for spotting this

PR #205, opened 2025-01-07
added application/octet-stream section

AtibQur:


sites/cheerpx/src/content/docs/11-guides/File-System-support.md
```diff
@@ -69,6 +69,31 @@ Files in the WebDevice are accessed relative to the current page's URL. For exam
...
+### Using `application/octet-stream` for WebDevice directories
+
+To handle files from WebDevices properly, set default_type `application/octet-stream`. This ensures that files are treated as binary data. **Only apply this to WebDevice directories**, or it may cause issues like broken images or malfunctioning scripts.
+
+**Nginx configuration example**:
+
+```nginx
+http {
+    # Set default type to application/octet-stream for WebDevice directories
+    default_type application/octet-stream;
```
alexp-sssup, line 81: This example should be structured differently. The file should be identical to the default configuration, and inside there must be a `location` section specific for the WebDevice that set the default mime type _and_ remove all mime type definitions by using `types { }`

We should also add a example line that creates the WebDevice to show that the path is the same used in the `location` section

alexp-sssup (changes requested):

GabrielaReyna:
Please ping me when Alessandro's suggested changes have been applied and I'll take a look. Thank you!

sites/cheerpx/src/content/docs/11-guides/File-System-support.md
```diff
@@ -69,6 +69,69 @@ Files in the WebDevice are accessed relative to the current page's URL. For exam
...
+            add_header 'Cross-Origin-Embedder-Policy' 'require-corp' always;
+        }
+
+        location /webdevice/ {
+            root .;
+            autoindex on;
+            types { }
+            default_type application/octet-stream;
+            add_header 'Cross-Origin-Opener-Policy' 'same-origin' always;
+            add_header 'Cross-Origin-Embedder-Policy' 'require-corp' always;
```
alexp-sssup, line 124: Can we add the headers only once outside of the location blocks, or by nesting location blocks? Unsure if that is supported or not, but let's try to avoid duplication if we can

sites/cheerpx/src/content/docs/11-guides/File-System-support.md
```diff
@@ -69,6 +69,69 @@ Files in the WebDevice are accessed relative to the current page's URL. For exam
...
+        }
+
+    }
+}
+```
+
+To verify it's working, simply navigate to `http://localhost:8080/webdevice/` in your browser.
```
alexp-sssup, line 131: This does not verify much, you get the autoindex which is the same thing you would get before configuring the mime types. A complete CX example would be more valuable here, but it might be too long for this section.

alexp-sssup (changes requested):

sites/cheerpx/src/content/docs/11-guides/File-System-support.md
AtibQur: Since we're going to add the references with all examples, I think that would be a nice place to add a complete example for it

sites/cheerpx/src/content/docs/11-guides/File-System-support.md
```diff
@@ -69,6 +69,65 @@ Files in the WebDevice are accessed relative to the current page's URL. For exam
 > [!note] Note
 > It's important to note that this behavior depends on the current page's URL, as it uses a relative path. For more predictable results, it's recommended to use absolute paths when possible.
 
+### Using `application/octet-stream` for WebDevice directories
+
+To handle files from WebDevices properly, set default_type `application/octet-stream`. This ensures that files are treated as binary data. **Only apply this to WebDevice directories**, or it may cause issues with image loading and JavaScript execution.
```
alexp-sssup, line 74:
```suggestion
To handle files from WebDevices properly, make sure the `application/octet-stream` MIME type is applied. This ensures that files are treated as binary data. **Only apply this to WebDevice directories**, or it may cause issues with image loading and JavaScript execution.
```

The purpose of this change is to tell the user _what_ is going on, before telling him _how_

sites/cheerpx/src/content/docs/11-guides/File-System-support.md
```diff
@@ -69,6 +69,65 @@ Files in the WebDevice are accessed relative to the current page's URL. For exam
...
+}
+
+http {
+    default_type  application/octet-stream;
+    access_log  nginx_access.log;
+    error_log   nginx_error.log info;
+
+    sendfile        on;
+
+	add_header 'Cross-Origin-Opener-Policy' 'same-origin' always;
```
alexp-sssup, line 102: Check indentation here

sites/cheerpx/src/content/docs/11-guides/File-System-support.md
```diff
@@ -69,6 +69,65 @@ Files in the WebDevice are accessed relative to the current page's URL. For exam
...
+	add_header 'Cross-Origin-Embedder-Policy' 'require-corp' always;
+
+    server {
+        listen       8080;
+        server_name  localhost;
+
+        gzip on;
+        gzip_types application/javascript application/wasm text/plain application/octet-stream;
```
alexp-sssup, line 110: This is wrong, we need to make sure compression is _not_ applied to these otherwise HTTP byte ranges won't work.

alexp-sssup (changes requested):

alexp-sssup (approved):

alexp-sssup:
@GabrielaReyna LGTM

GabrielaReyna (approved):

PR #221, opened 2025-02-19
Add missing step cheerpx networking docs

YoussefHenna:
Hey, just a quick docs update for something that took me a while to find.

The `networkLogin()` call to trigger the tailscale connection was undocumented and unclear to me that the network connection was not going to happen automatically. Adding this to the docs will probably help others who run into the same issue.


alexp-sssup (approved):

alexp-sssup:
Thanks for your help, this step was indeed missing.

PR #228, opened 2025-03-14
Claude: Migrate post

alexp-sssup:


content/blog/webvm-claude.mdx
```diff
@@ -0,0 +1,346 @@
...
+## Pointers, keys and (lots of) screenshots
+
+Contrary to the **bash** tool, the **computer** tool has many more moving parts, but it’s also much more capable... when things go well. This feature depends heavily on screenshots to “see” the screen and decide how to interact with it. Although this makes it possible for the model to have full understanding of the environment, it comes with some downsides:
+
+- **Cost**: Since every interaction depends on one or more screenshots, which consumes lots of tokens, conversations can get quite expensive
+- **Reliability**: The precision of the model when clicking on UI elements is not amazing. This might be connected to the maximal resolution of the screenshot being [capped to 1024x768](https://docs.anthropic.com/en/docs/agents-and-tools/computer-use#computer-tool), requiring rescaling of the screenshots and mouse coordinates to compensate for this.
+
+While the **bash** tool only had one single action that could be used, the **computer** tool has several. We'll describe a few of them in depth, see [here](https://docs.anthropic.com/en/docs/agents-and-tools/computer-use#computer-tool) for a complete list.
+
+### The “screenshot” action
```
GabrielaReyna, line 183: Please use single word/line code snippet on titles and subtitles instead of double commas.
```suggestion
### The `screenshot` action
```

content/blog/webvm-claude.mdx
```diff
@@ -0,0 +1,346 @@
...
+		var imageSrc = {type: "base64", media_type: "image/png", data: dataUrl};
+		var contentObj = {type: "image", source: imageSrc};
+		return [contentObj];
+	}
+}
+```
+
+This is the [actual code](https://github.com/leaningtech/webvm/blob/91f68ce044d35489eaebf1703da3c408787c9c88/src/lib/anthropic.js#L288) used in WebVM, lightly edited for clarity. The inline comments should make it fairly easy to follow. The general idea is to keep the original display at native resolution in its `<canvas>` element, and use a secondary `<canvas>` element to get screenshots, using the `toDataURL` API. The original display is copied and resized into the secondary `<canvas>` using the `drawImage` API. To avoid incurring excessive costs, we keep track of the current state of the display and only send a reply when something has actually changed. This used to be very important with Claude 3.5, but may not be necessary with Claude 3.7 thanks to the newly introduced “wait” action.
+
+### The “mouse_move”, “left_click” and “right_click” actions
```
GabrielaReyna, line 236:
```suggestion
### The `mouse_move`, `left_click` and `right_click` actions
```

content/blog/webvm-claude.mdx
```diff
@@ -0,0 +1,346 @@
...
+	display.dispatchEvent(me);
+	return null;
+}
+```
+
+The implementation currently sends simulated mouse events to the display: this is suboptimal and in the future we plan to expose a direct CheerpX API to inject events without depending on the browser event loop.
+
+An important detail is the 60 millisecond delay between the `mousedown` and `mouseup` events. It is required to make sure the Xorg process does not ignore the `mouseup` due to its _debouncing_ logic. If the delay is too short the `mouseup` will be considered a hardware glitch, rather than valid user input.
+
+### The “type” and “key” actions
```
GabrielaReyna, line 264:
```suggestion
### The `type` and `key` actions
```

content/blog/webvm-claude.mdx
```diff
@@ -0,0 +1,346 @@
...
+               f(null);
+       });
+}
+```
+
+Similarly to the mouse input above, the `kmsSendChar` helper function is currently also based on simulating KeyboardEvents. We plan to simplify this in the future by providing CheerpX APIs for this case as well.
+
+Implementing the **key** action requires some (not truly interesting) additional parsing to support the expected `xdotool` syntax, typical inputs could be “ctrl+C”, “Return”, “Escape”. Directly executing `xdotool` in the VM would have been another option, but we preferred keeping the logic in JavaScript for educational purposes and to be compatible with future non-Linux or non-Xorg based VMs.
+
+### The “wait” action
```
GabrielaReyna, line 293:
```suggestion
### The `wait` action
```

GabrielaReyna (changes requested):

content/blog/webvm-claude.mdx
alexp-sssup: I've considered the use backticks, but I suspect "code" formatting in the middle of heading will be off.

GabrielaReyna (approved):

PR #262, opened 2025-06-02
Added pagefind filter for language searching

Orpheus-3145:
Changes made:

- passing current locale to pagefind svelte component for filtering
- added tag `<div>` with pagefind attribute for custom language filtering in every file in _src/content/docs_


alexp-sssup:
Might I advice to consider english to be the "default"? So that we can avoid adding the tag everywhere all the time? We are guaranteed to forget about it

GabrielaReyna:
Hi Francesco, thanks for working on this, seems like you have found a workaround. 
Even when the solution works, I think adding the html tag may be a bit troublesome in the future if the language is not part of the overall structure of the content collection. 

The potential solution to not forget about adding the english language tag and also making it part of the content collection structure is to add it as a non optional attribute on the content collection for docs ([see packages/astro-theme/content.ts](https://github.com/leaningtech/labs/blob/main/packages/astro-theme/content.ts))

something like:
`language: z.string().optional(),`

We might need to give two options: `en` or `ja`.  Unsure if we can give a default value if it is not defined, we could look into it. If it has to be defined, it would not pass the check because the .md attributes are not complete.

In terms of how to filter an attribute of the content collection, you can take a look on how some other filters in labs work: 
- Filter if a blog post is "featured" (important) to showcase at the bottom of l[abs landing page](https://labs.leaningtech.com/): see the component [BlogPostCardset.astro](https://github.com/leaningtech/labs/blob/main/packages/astro-theme/components/BlogPostCardSet.astro). 
- Filter the blog posts shown on each product specific blog (e.g. cheerpx.io/blog):[ See how the blog index filters by product name](https://github.com/leaningtech/labs/blob/main/packages/astro-theme/pages/blog/index.astro). The product name is an attribute name `tags` on the blog post markdown files as shown on the[ content collection file above](https://github.com/leaningtech/labs/blob/main/packages/astro-theme/content.ts).

PR #264, opened 2025-06-10
CheerpJ filesystem tutorial

theodoravraimakis:
Filesystem tutorial


alexp-sssup:
In terms of which repo to use we should try to avoid making many new ones, already too many require manual synchronization alerady. Let's put everything in cheerpj-meta

sites/cheerpj/src/content/docs/13-tutorials/06-filesystem.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+In the second part, we'll explore a Java Applet that can also perform file downloads using different methods, including `showDocument()` for direct URL downloads and a native method for more complex interactions. This will involve `AppletExample/AppletExample.java` and `applet.html`.
+
+The precompiled JAR files (`FilesystemExample.jar` and `AppletExample.jar`) are provided for convenience, but you can recompile the Java code using the `Makefile` if you make any changes.
+
+## CheerpJ's Virtual Filesystem
+
+CheerpJ's virtual filesystem has several key mounting points that behave differently:
+
+- **`/files/`**: This is a **temporary, in-memory filesystem**. Files written here are accessible by your Java application but are not persistent across sessions or page reloads. This is useful for temporary files or data generated during a single user session that needs to be downloaded.
```
epanholz, line 44: This is wrong, the /files mounting point is persistent and will only be cleared if the user clears out indexdb manually. Our own docs explain it -> https://cheerpj.com/docs/guides/File-System-support#files-mount-point

epanholz:
@GabrielaReyna I agree that this content makes more sense as a guide, but I think it should be it's own guide. The existing one mostly explains the architecture and briefly mentions some APIs that can be used to interact with the virtual file system. We could rename the existing guide to `Files and Filesystem Architecture` and call this one `Files and Filesystem Usage Guide`. Otherwise I am afraid we might create one very large docs page that is hard to follow. How does that sound?

theodoravraimakis:
Hi, I have updated the files as per your last comments @epanholz @GabrielaReyna

epanholz (approved):
This is now ready for merging from my side
@GabrielaReyna please go ahead and merge it when you can

PR #279, opened 2025-09-16
Add name and version to package.json

Ceriumin:
Support for installing through pnpm as it requires a "name" to be present in order to install


Ceriumin:
Just imported the package for the global navbar. So weirdly enough npm lets you install it without issues but pnpm throws an error saying the package.json is missing because it doesn't have a name, very odd. Either way it works now, thanks

PR #292, opened 2025-10-23
Standardize local hosting instructions in Getting Started and Tutorials

rijulshrestha:
- Standardize to npx http-server -p 8080 across Getting Started and Tutorials.
- Standardized local hosting instructions.


epanholz (approved):
This looks good to me.
Context for @GabrielaReyna, Rijul noticed that we used different HTTP server one liners throughout the docs so we decided to update them and use the same everywhere for consistency.

PR #323, opened 2025-12-05
Remove hackathon buttons from overview and hero

jadentel:
removed the hackathon invitation buttons inside cheerpj docs/overview and landing page of labs homepage.


alexp-sssup:
For the sake of efficiency I went ahead and fixed the remaining linter issues.

PR #330, opened 2025-12-10
FlatLaf Demo

rijulshrestha:


GabrielaReyna (changes requested):
Hi Rijul,

thanks for the PR!
Any CheerpJ demos should be contained in the demos repo: https://github.com/leaningtech/cheerpj-demos

Unless there is a specific reason you are doing it in Labs, please move it there when you get the chance. 

Many thanks!
Ana

GabrielaReyna:
Thanks for clarifying!

PR #374, opened 2026-03-26
File system guides updated for the new upload/download functionality

rijulshrestha:


epanholz (changes requested):
Looking good already. Requested some more small changes.

epanholz (changes requested):
I added one last comment about the table in the explanation. Once that is applied it's ready for merging. Good job.

**Note:** I will be holding off merging this until we release a new point version of CJ, which we plan to do soon.

GabrielaReyna (changes requested):
Looks good. As a minor thing, the directory tree could be added as a code block, just like we did in http://localhost:4321/docs/tutorials/swingset3 at the beginning of both guides.

GabrielaReyna (approved):

epanholz (approved):

PR #381, opened 2026-04-13
Submission

jadentel:
created a submission page for the hackathon.

### DIFF
--- a/packages/astro-theme/components/hackathon/HackathonSubmission.astro
+++ b/packages/astro-theme/components/hackathon/HackathonSubmission.astro
@@ -1,0 +1,340 @@
+---
+interface Props {
+	eventName: string;
+	endpoint: string;
+	opensAt: string;
+	closesAt: string;
+}
+
+const { eventName, endpoint, opensAt, closesAt } = Astro.props;
+---
+
+<div
+	class="min-h-screen bg-[#050505] flex items-center justify-center px-6 py-20"
+>
+	<div class="w-full max-w-4xl grid md:grid-cols-2 gap-12 items-center">
+		<!-- Left -->
+		<div class="space-y-5">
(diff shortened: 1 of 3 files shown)


GabrielaReyna (changes requested):
Hi Jaden,

Great work! I like the visuals and cleanliness of the form. 

I have some suggestions to make this reusable for future events. 
- Make the page a component under the /packages/astro-theme/components. You can create a folder called "hackathon" and put it there (and other components we make in the future around hackathons). Then, create a page under labs while importing the component from the package. 
- Please also gather the discord username, and allow a field for team leader and the users of the rest of the team. 
- The component itself can receive arguments (props). This ensures re-usability and is not just hard-coded. Please consider as props: EventName, the sheet endpoint url, time and date of submissions opening, time and date of submissions closing. 

About the opening and closing of submissions: feel creative to put a countdown or a custom message when submissions are closed. Since this is client-side, make sure that the endpoint script stops receiving entries at the time we set. 

We can do a test once you work on it :)

codingfrog27:
Hi Team!

Like Gaby said, looks nice Jaden! 


## My testing
- I tested both earlier and later time submissions and it seems to work! 
- But Like Gaby said the server side should also have a close time set to protect against people sending a json to it directly. I didn't have access to the google script so I'm not able to check if you added this already.
- I fished my id out of the response and the edit url also seems to work! Though doing it this way manually. I did


For now I'd say **We can merge this branch** :)


## Future feature suggestions

I know this was made as more as a proof of concept, but for future use I think we should add:

- Returning the ID to the user and telling them to save it for submission editting, as well as adding a button where they can fill it in to access said page
- I like the minimalism but think we can add a bit more customization! Ie:
     - A banner image path that we pass as prop to the component (and optionally a default fallback banner image)
     -  Props for custom messages, like a welcoming paragraph

GabrielaReyna (approved):

PR #382, opened 2026-04-14
Update svelte-browserpod-editor and load browserpod from the CDN dire…

yuri91:
…ctly

This way we can load different browserpod versions for different articles/demos


packages/astro-theme/components/blog/BrowserPodDemo_1.0.svelte
```diff
@@ -10,21 +11,28 @@
...
 		{
 			id: "build",
 			label: "Build",
-			commands: [
-				["npm", "install"],
-				["npm", "run", "dev"],
-			],
-			autoRun: true,
+			onReady: async (run) => {
+				console.log("ready");
```
alexp-sssup: This logging might be not intended

alexp-sssup (changes requested):

packages/astro-theme/components/blog/BrowserPodDemo_1.0.svelte
yuri91: good catch

PR #386, opened 2026-04-16
Using ES6 module import in both code snippets instead of direct script include

smclacke:
Removed ES6 compatibility info since using ES6 module in example code makes it redundant information. This was discussed with Jules.

### DIFF
--- a/sites/cheerpx/src/content/docs/10-getting-started/index.md
+++ b/sites/cheerpx/src/content/docs/10-getting-started/index.md
@@ -30,7 +30,7 @@ The example below demonstrates how to set up the file system and devices using [
 		<meta charset="utf-8" />
 		<title>CheerpX Getting Started</title>
 		<script type="module">
-			import * as CheerpX from "https://cxrtnc.leaningtech.com/%CX_LATEST%/cx.js";
+			import * as CheerpX from "https://cxrtnc.leaningtech.com/%CX_LATEST%/cx.esm.js";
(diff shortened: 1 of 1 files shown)


sites/cheerpx/src/content/docs/10-getting-started/index.md
```diff
@@ -43,8 +29,8 @@ The example below demonstrates how to set up the file system and devices using [
 	<head>
 		<meta charset="utf-8" />
 		<title>CheerpX Getting Started</title>
-		<script src="https://cxrtnc.leaningtech.com/%CX_LATEST%/cx.js"></script>
 		<script type="module">
+			import * as CheerpX from "https://cxrtnc.leaningtech.com/%CX_LATEST%/cx.js";
```
yuri91: I think we should instead use the ESM version of the script:

```
import  CheerpX from "https://cxrtnc.leaningtech.com/%CX_LATEST%/cx.esm.js";
```

smclacke: thanks!

smclacke: as far as I can see 'import CheerpX' only imports default exports, so 'import * from CheerpX' is necessary. My latest push updates both scripts to have 'cx.esm.js' extension which is indeed correct.

yuri91 (approved):

PR #389, opened 2026-04-16
Single showcase and blog

GabrielaReyna:
Please revise that everything works with care:
- Blog, blog authors and showcase content only exist in labs
- Not copying files needed anymore
- pages for showcase and blog only in labs
- common collections now are only docs, labs required its own collection (to add blog and showcase)


alexp-sssup:
I must be missing something here. it seems to me this PR wipes out the blog completely

GabrielaReyna:
> I must be missing something here. it seems to me this PR wipes out the blog completely

oopsis, I might have deleted the content folder by accident. I will take a look.

packages/astro-theme/components/nav/product/Nav.astro
```diff
@@ -23,12 +23,8 @@ const firstDocsEntry = getLocalisedProductNav(
 	Astro.currentLocale
 )[0];
 if (!firstDocsEntry) throw new Error("must have a docs entry");
-let blogref = "/blog";
-let showcaseref = "/showcase";
-if (product.id.includes("cheerpj") || product.id.includes("browserpod")) {
-	blogref = "/docs/blog";
-	showcaseref = "/docs/showcase";
-}
+let blogref = "https://labs.leaningtech.com/blog";
```
codingfrog27, line 26: The only downside to this is that now any local build links will still point to the live version. I wonder if we can use an env variable or other helper to detect wether we're in local development or not (but can be done in future PR as well)

codingfrog27 (approved):
Looks good to me! I found and ammended a small regression (potentially also caused when solving merge issues) where Elisabeths author picture reverted to an older one (both were present, my guess is the other location takes priority). 

Nice work and excited for the build time decrease :)

PR #392, opened 2026-04-20
Fixed CJ docs build failing

codingfrog27:
CJ/docs was failing because the new showcase branch didn't have a translation for the new demo link on the navbar. Should work now.

Translation is AI made so might want to check with a Native speaker


alexp-sssup:
@GabrielaReyna Do you think we can have the translation quickly checked? This issue is currently blocking the CJ 4.3 release.

GabrielaReyna:
Translation is correct

PR #396, opened 2026-04-22
cx quick start update

smclacke:
missed a script include in the quick start, fixed that and updated a few sentences so it is consistent with module import


GabrielaReyna (approved):
Good catch with the import!

PR #403, opened 2026-04-29
Labs build warning fixes and optimisations

codingfrog27:
# Build fixes:
A multitude of small build optimisations and warning fixes/silencing.

- **Tailwind pattern:** adjusted tailwind content scan pattern (halves build time)
- **Failed to resolve font file warning** put the font in an inline style block to prevent vite from pre resolving it and adding /docs to the path, causing it to fail.
   - Also removed the bracketed name file extention in the name for the font file as we seem to only use woff2 and it could cause issues.
- **Chunk size warning:** Vite warns for a couple of chunks that are larger than 500kb.. I looked into some options to make these smaller but the large chunks were mostly demos. Opted to increase the warning size to 700 kb instead. (Can see the argument that the warning is better to keep.. Some demos do still pass over 700kb anyways. I'd say 500 is acceptable but more than 700 is worth seeing, hence this number)
- **npm unkown env config warning**:  The warning was caused by us running npx pagefind, I replaced it with a local binary (Which should work on both Unix and windows). This did feel a bit off but from what I've read it's standard practice? Feedback welcome.
- **Added HTML wrapper for index.astro:**  Pagefind gave a warning pages were missing the html element, fix should also help with SEO optimisation.
## What to test
- using the search bar on windows machines
- Deeper testing that no UI elements got removed with the adjustment of the tailwind pattern. So far I've only checked Browserpod docs.

## Future fixes
- Centralize the font into a dedicated asset for caching optimisation (Either by putting it in theme/root head or via @font-face)
- Look into if we can lazy load the demos causing the large chunks

## For Merging!!
Since these are a lot of technical changes I think we shouldn't squash it all into 1 commit Ive cleaned up my commit history to have one fix per commit. So please rebase onto main rather than the usual squash and merge. :)


GabrielaReyna:
If I remember correctly, Alessandro once pointed out that the green dot animation next to the discord icon on the documentation pages slows down the performance. If you have the capacity to include that fix in this PR, it would be great.

codingfrog27:
re: removing the discord CSS animation. This was my plan but Alessandro advised me to do so in a separate PR, to prevent this from going stale/

GabrielaReyna (approved):

PR #405, opened 2026-05-01
Swing tutorial jar comment

smclacke:
Added a comment to the swingset3 tutorial to clarify the app we use causes harmless errors.


GabrielaReyna:
I am taking a look to this now. I might come with further instructions.

GabrielaReyna (changes requested):
Hi Sarah, thanks for taking a look at this. 

I'd like for you to do the following in this PR:
- I believe some of the jars in the swingset tutorial are empty and not needed. Please test this out and get rid of unnecessary files. These changes should apply to the zip file in the tutorial, the examples folder in the cheerpj-meta repo and in the swingset3 leaning tech repo ( the one deployed and used in the tutorial iframe). If you have checked all of this, feel free to just ignore this bullet point. 
- The zip files in all the tutorials of cheerpj contain the HTML file where cheerpj is loaded via the loader URL. This is a pain to update every single time there is a new version, zip back and modify the tutorials. Hence, please modify the tutorial's HTML file and replace the loader URL with "paste url here" maybe pointing to a url in the docs with the current loader URL. Please do for all the cheerpj tutorials that have a downloadable layout.
- Once the bullet point above is complete, please mention in the tutorials that the url needs to be updated. 


If you have a better suggestion on how to do the above, for example, automating creating a zip that exists in the examples repo rather than in the labs repo would be amazing. But if not, the simple solution above should be enough.

Many thanks!

alexp-sssup:
Apologies for the late reply here, but I think we should get rid of these zip files altogether in favor of using git.

In terms a usability a pre-made zip file with a invalid URL will be confusing.

alexp-sssup:
Another thing I just realized: GitHub itself provides the functionality to download a repo as ZIP. At least IIRC.

GabrielaReyna:
> Another thing I just realized: GitHub itself provides the functionality to download a repo as ZIP. At least IIRC.

It would download the entire repo, no the subdirectory where the specific example is. 

We have different options here:
- We get rid of the zip files and ask the reader to simply download the necessary jars from the repo, manually.
- We ask the user to clone/download the examples repo and instruct which folder to use (the one of the specific example)
- or we can download the specific example by creating an Astro component that runs client-side (download the entire repo, and then packs the specific subfolder), with the limitations of the GitHub API (60 requests per hour per IP when unauthenticated)

alexp-sssup:
@smclacke Heads up, you'll need to rebase your branch on `main`, otherwise the commit list will include spurious, already merged, work

GabrielaReyna:
Your branch is now rebased onto main; you should see only your changes in this PR.

Please continue and let me know when it is ready for review :)

sites/cheerpj/src/content/docs/13-tutorials/00-swingset3.mdx
```diff
@@ -11,8 +11,18 @@ In this tutorial, we'll run the SwingSet3 application in the browser.
...
+	git clone https://github.com/leaningtech/cheerpj-meta.git
+```
+
+- Navigate to the SwingSet3 tutorial directory:
+
+```sh
+	cd examples/SwingSet3
+```
 
 The starting point of this example is an empty HTML page, the SwingSet3 jar, and its dependencies:
```
GabrielaReyna: This example template (in the repo) doesn't have an empty HTML file, so it would be a good idea to mention that the HTML file needs to be started from scratch, just as you did in the applet tutorial.

GabrielaReyna (changes requested):
Just a tiny change requested, once amended, this PR is ready to merge.

GabrielaReyna (approved):

PR #406, opened 2026-05-01
Corrected Datadevice being writable in the Filesystem support table

codingfrog27:


codingfrog27:
Will merge since it's a one word change

PR #414, opened 2026-05-18
further blog banner image resolution increase

codingfrog27:
increased resolution and added logic to height.

Resolution had been increased previously but still needed further increasing to avoid bluriness.

Besides this, the height resolution did not have the same check the width had to check for the aspect ratio, causing a mismatched aspect ratio that astro would then process to make it still fit, resulting in more bluriness.


GabrielaReyna (commented):
Thank you! This is for the blog post card component, not the blog post banner. 

For the banner, we need to take a look at `packages/astro-theme/layouts/BlogPost.astro`

PR #418, opened 2026-06-02
Added sitemap to BrowserPod docs to link with commercial site, correct seo issues and remove SERP indexing errors

Liamoscarmccaffrey:


alexp-sssup:
@GabrielaReyna Please double check that this solution actually works with a static build. I am not convinced the SSR pipeline will do the right thing on an API route

GabrielaReyna (approved):
Hi

I built the site, and sitemap.xml is created with the content of the docs pages, everything starts with browserpod.io/docs so seems to work well. 

@alexp-sssup I think this resolves your concern.

alexp-sssup:
If `sitemap.xml` is generated then we should be in the clear. Please go ahead and merge

PR #419, opened 2026-06-02
added gtm tag to browserpod docs

Liamoscarmccaffrey:


packages/astro-theme/components/BaseHead.astro
```diff
@@ -100,6 +100,13 @@ const localised = getLocalisedPaths(Astro.url.pathname, Astro.currentLocale);
 	))
 }
 
+{
+	product?.id === "browserpod" && (
+		<!-- Google Tag Manager -->
+		<script is:inline>{`(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-M4T2J9TS');`}</script>
```
GabrielaReyna: The is:inline script tag should contain the JS directly, not wrapped in backticks inside {``}. The browser sees the backticks and curly braces as a string, not code. Hence GMT is never loaded.

Please change` <script is:inline>{`(function...)`}</script>` to `<script is:inline>(function...)</script>`

GabrielaReyna (changes requested):
Please also do a pass with prettier `pnpm prettier . --check` and  `pnpm prettier . --write`

If you want I can do all these changes for you, just let me know.

GabrielaReyna (approved):

PR #420, opened 2026-06-03
Cx image building doc corrections

codingfrog27:
Corrected the image building guide and full OS tutorial to use mounting in the unshare namespace to avoid permission issues. While I was at it I made some changes to improve the general flow and structure of related docs.

- Added a step to check filesystem size and to add a bit of extra space (but also to not exceed 2GB)
- Put the optional edit /etc/subgid step behind the conclusion as it won't be relavent for most users. But also referenced it at the mounting step for the cases where it would give issues
- Added prerequisites and a small docker intro note
- Made the tutorial instead refer to the guide but just add the useradd lines to minimize duplicate documentation

## Changes to consider/feedback
- Would the prerequisites be too much? Think it's nice but it is dropping our consistency a bit since other guides don't have them (though again if i remember some already do so might not be much consistency there anyways) 
- Some steps, mainly the part where we mount the file could be condensed with some inline commands, this would also save the user from having to copy and paste the path to their local mountpoint. But I opted for more easily understandable steps
- I do at this point have a script users we could include for users to easily build their image themselves. But the risk here would be more of them coming to us or giving up if it doesn't work and us having to potentially maintain more if anything changes in the future.


sites/cheerpx/src/content/docs/11-guides/custom-images.md
```diff
@@ -3,84 +3,109 @@ title: Custom disk images
 description: How to create custom ext2 images to use as a filesystem for CheerpX
 ---
 
-This guide will show you how to configure a Docker container and build it into an `ext2` image for use with CheerpX.
+This guide will show you how to configure a container image and build it into an `ext2` image for use with CheerpX.
```
saarikoski-jules: container image feels like confusing terminology, as they're, in docker terminology, different things. I understand getting rid of the word "docker" though. Update the terminology everywhere else wehre its used as well.

sites/cheerpx/src/content/docs/11-guides/custom-images.md
```diff
@@ -3,84 +3,109 @@ title: Custom disk images
...
 
-Once the `Dockerfile` is ready, build the image. Use `buildah` to create a container image for your filesystem.
+Once the `Dockerfile` is ready, use `buildah` to create a container image.
 
 ```bash
 buildah build -f Dockerfile --dns=none --platform linux/i386 -t cheerpximage
 ```
 
-## 3. Create a container from the Docker Image
+### 1.3 Create a container from the Docker Image
```
saarikoski-jules: If you choose to get rid of the Docker terminology, you should also remove it here (but feel free to decide otherwise)

sites/cheerpx/src/content/docs/11-guides/custom-images.md
```diff
@@ -3,84 +3,109 @@ title: Custom disk images
...
+## 2 Building our ext2 image
+
+With the container up and running, we can now use it's filesystem to build the actual `ext2` disk image that will be used by CheerpX.
+
+### 2.1 Enter a user namespace
+
+_This step can be skipped if running as root_
 
-Create a directory for the filesystem:
+Before we can use the containers filesystem, we will enter a user namespace to make sure the original file ownership stays intact.
```
saarikoski-jules: This isn't exactly true. We use unshare to get around doing things in the system, without becoming root. It's called "rootless mode". In reality what we want is for the user to not be able to destroy their system by not understanding what they're doing, and still be able to create an ext2 image.

sites/cheerpx/src/content/docs/11-guides/custom-images.md
```diff
@@ -3,84 +3,109 @@ title: Custom disk images
...
 ```
 
-## 4. Copy the Filesystem from the Container into a local directory
+## 2 Building our ext2 image
+
+With the container up and running, we can now use it's filesystem to build the actual `ext2` disk image that will be used by CheerpX.
+
+### 2.1 Enter a user namespace
+
+_This step can be skipped if running as root_
```
saarikoski-jules: We really don't want our users to be root. But it's fair to explain what the purpose of this is.

sites/cheerpx/src/content/docs/12-tutorials/full_os.md
```diff
@@ -8,41 +8,16 @@ This tutorial will guide you through creating a custom filesystem from scratch a
...
 
 We will create an Ext2 image, which will serve as the root filesystem for CheerpX.
+You can build the image by following the [Custom disk images][custom-disk-images] guide, but **Make sure these lines to the end of your dockerfile** so the image has a login user and a root password:
 
-To ensure consistency in preparing the image contents, we will use a Dockerfile, you can edit the Dockerfile to add custom packages or change the base to the distro of your choosing. It's important to select the `i386` architecture, since CheerpX does not currently support 64-bit executables.
-
-```dockerfile title=Dockerfile
-FROM --platform=i386 docker.io/i386/debian:buster
-ARG DEBIAN_FRONTEND=noninteractive
+```dockerfile
```
saarikoski-jules, line 13: This is much better, to not have to remember to always update everything

saarikoski-jules (changes requested):
The commands are good, it's what I use usually. Left a few small notes, but nothing significant.

About your questions:
When it comes to prerequisites, I'll let you make your own judgement call.
About the steps, I prefer clarity over conciseness. I think what you've done is good.
We do not want to give the users a script. There's a higher chance of people trying to understand what they're doing if we give them steps instead.

PR #421, opened 2026-06-04
Cj docs remove await cheerpOSAddStringFile()

smclacke:
- removed unnecessary await call for the sync cheerpOSAddStringFile() function
- checked rest of docs for other spurious awaits in front of sync funcs
- remove async + try catch block from code snippet in question since it's unnecessary and clutters the code snippet (ref: filesystem method 2)


codingfrog27 (approved):
Looks good to me! 

As we discussed the try and catch block removal is more subjective, but I agree with your argument that simplicity/conciseness is more valuable here.

PR #422, opened 2026-06-08
updated BaseHead with path-to-URL mapping only for the BrowserPod blogs

Liamoscarmccaffrey:
Ensuring blogs canonical point to the browserpod.io versions (i.e., in leaningtech.com repo).

### DIFF
--- a/packages/astro-theme/components/BaseHead.astro
+++ b/packages/astro-theme/components/BaseHead.astro
@@ -18,7 +18,24 @@ const isBlogPost =
 const canonicalSite = isBlogPost
 	? new URL("https://labs.leaningtech.com")
 	: Astro.site;
-const canonicalURL = new URL(Astro.url.pathname, canonicalSite);
+
+// BrowserPod posts that have been migrated to browserpod.io point their
+// canonical at the browserpod.io equivalent to avoid duplicate-content conflicts.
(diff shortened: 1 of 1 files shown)


alexp-sssup (approved):

alexp-sssup:
I've checked this and this seems ok, but the maintainability is very poor. A better approach would be to do this for any post tagged browserpod. This said, the whole duplicated content is a maintenance nightmare.

alexp-sssup:
@GabrielaReyna I'd like your opinion before merging

GabrielaReyna (approved):
Hey! Found two bugs in the BrowserPod canonicals PR during testing:

`.html` suffix not stripped — the preview deployment appends .html to pathnames (e.g. /blog/browserpod-deep-dive.html), so the canonical lookup was falling through to the labs fallback instead of pointing at browserpod.io. Fixed by updating the normalisation regex to strip `.html` as well as trailing slashes.
Wrong map keys for 1.0 and 2.0 posts — the lookup table used `/blog/browserpod-1.0` and `/blog/browserpod-2.0` as keys, but the actual paths on the site are `/blog/browserpod-10` and `/blog/browserpod-20` (no dot). Those entries were never matching. Fixed by correcting the keys.
Both fixes are on the branch now.

PR #425, opened 2026-06-16
Fix: Updated broken links that were registering 404s due to issues with majuscules and slashes

Liamoscarmccaffrey:
Did some digging and found the root cause of the 404 errors, which would not be fixed with the sitemap. Please see the changes, hopefully they're okay. It's claude heavy.


GabrielaReyna:
For awareness, I started reviewing this today. I found a couple of bugs. I need to continue testing and will push the changes tomorrow.

GabrielaReyna (approved):

PR #428, opened 2026-06-22
CJ doc update debug mode

smclacke:
added a notice about debug mode affecting performance in three applicable places in the docs

### DIFF
--- a/sites/cheerpj/src/content/docs/00-overview.mdx
+++ b/sites/cheerpj/src/content/docs/00-overview.mdx
@@ -71,6 +71,8 @@ With CheerpJ, you can:
 
 **No installation required:** CheerpJ runs entirely in the browser, so there’s no need to download or install additional software.
 
+**Debug mode:** See [Debugging CheerpJ] for details. Enabling debug mode can significantly affect performance, disable it when not needed.
+
 CheerpJ is based on a full OpenJDK runtime environment and supports:
 
 - [Networking]
@@ -187,3 +189,4 @@ We have a vibrant Discord community where you can ask questions, share your proj
 [Networking]: /docs/guides/Networking
 [Virtualized filesystem]: /docs/guides/File-System-support
 [Clipboard]: /docs/reference/cheerpjInit#clipboardmode
+[Debugging CheerpJ]: /docs/guides/cheerpj-debug
--- a/sites/cheerpj/src/content/docs/00-overview.mdx
+++ b/sites/cheerpj/src/content/docs/00-overview.mdx
@@ -9,7 +9,6 @@ import ShowcaseList from "@leaningtech/astro-theme/components/ShowcaseList.astro
 import { DISCORD_URL } from "@/consts.ts";
 import JNLPRunnerButton from "@leaningtech/astro-theme/components/JNLPRunnerButton.astro";
 import AppletRunnerButton from "@leaningtech/astro-theme/components/AppletRunnerButton.astro";
-import Callout from "@leaningtech/astro-theme/components/Callout.astro";
 
 <div class="not-prose flex gap-2 mb-2">
 	<img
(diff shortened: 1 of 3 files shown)


sites/cheerpj/src/content/docs/00-overview.mdx
```diff
@@ -71,6 +71,8 @@ With CheerpJ, you can:
 
 **No installation required:** CheerpJ runs entirely in the browser, so there’s no need to download or install additional software.
 
+**Debug mode:** See [Debugging CheerpJ] for details. Enabling debug mode can significantly affect performance, disable it when not needed.
```
alexp-sssup: I feel this is not the right place for this information. It's not a top-level "feature"

alexp-sssup (changes requested):

sites/cheerpj/src/content/docs/00-overview.mdx
smclacke: yes, I agree. I've moved this to a Tip box format under the tutorial section, I think this makes more sense

sites/cheerpj/src/content/docs/00-overview.mdx
```diff
@@ -114,6 +114,9 @@ Know what you're building? Jump straight to the relevant tutorial:
 	</a>
 </div>
 
+> [!TIP] Debug mode
+> See Debugging CheerpJ for details. Enabling debug mode can significantly affect performance, disable it when not needed.
```
alexp-sssup: Maybe we do want a section link here?

alexp-sssup (changes requested):

sites/cheerpj/src/content/docs/00-overview.mdx
smclacke: I'll look more into it, my three different approaches to make this a link have so far been unsuccessful.

sites/cheerpj/src/content/docs/00-overview.mdx
```diff
@@ -114,6 +115,12 @@ Know what you're building? Jump straight to the relevant tutorial:
 	</a>
 </div>
 
+<Callout variant="tip" title="Debug mode">
```
alexp-sssup: This seems ok to me, but I am not convinced that this is the right place for this information. Anyway, I'll need @GabrielaReyna complete the review, it's ok for the rest from my perspective

GabrielaReyna: I agree this is not the right place and probably some quick restructuring of the page is needed here in order to fit it. 
- The "not a developer?" subheading is spoiling the table of contents indexing and maybe should be moved down, and make the subheading the same level as the rest.
- Make a subsection to fit the debugging guide link, for example "Debugging CheerpJ" and a 1 short paragraph invite to the tutorial instead of a callout

Hence the table of content should look something like this:
 - Features
- Getting started
- How does CheerpJ work?
- What's Unique about CheerpJ?
- Debugging CheerpJ
- Licensing
- Demos
- Not a Developer?
- Community

sites/cheerpj/src/content/docs/11-guides/cheerpj-debug.mdx
```diff
@@ -7,6 +7,9 @@ import { DISCORD_URL } from "@/consts.ts";
 
 Sometimes, you might encounter issues or experience unexpected behavior when using CheerpJ. Don’t worry if that happens — this guide will help you troubleshoot the problem and gather essential resources, in case you need further assistance with debugging.
 
+> [!important] Heads up
```
GabrielaReyna: This feels a bit out of context since maybe the user doesn't know what "Debug Mode" is, seems like this callout fits better under the `[Enable advanced debug logs in CheerpJ](http://localhost:4322/docs/guides/cheerpj-debug#enable-advanced-debug-logs-in-cheerpj)` section

GabrielaReyna (changes requested):

sites/cheerpj/src/content/docs/00-overview.mdx
smclacke: I wasn't really sure what other information to put under 'debugging cheerpj' since it would become a repeat of what is actually in the debugging document. Let me know if you think I should add more.

sites/cheerpj/src/content/docs/11-guides/cheerpj-debug.mdx
GabrielaReyna:
```suggestion
[`enableDebug`]: /docs/reference/cheerpjInit#enabledebug
```

GabrielaReyna: url is uppercase sensitive so this link doesn't actually take you to the option, but to the general page

GabrielaReyna (approved):

PR #432, opened 2026-07-01
Add hackathon winners to showcase

jadentel:


smclacke (approved):
This all looks really nice to me! I only wonder why we don't give credits, as in naming the individuals? Maybe something I'm not aware of or was decided against. Nice work!

GabrielaReyna:
> This all looks really nice to me! I only wonder why we don't give credits, as in naming the individuals? Maybe something I'm not aware of or was decided against. Nice work!

Jaden actually asked me if he should add the authors' names in the authors tag, but probably it would get too crowded. Alternatively, we could add the team member's names in the description. something like "DevHub team was composed of name1, name2, name3"

smclacke:
I like the way you've added the credits without making anything too crowded. Nice job, my approval stands ;)

PR #436, opened 2026-07-16
Feat: Added Google Tag Manager to CheerpJ docs.

Liamoscarmccaffrey:
This pull requests adds the Google Tag Manager tag to the CheerpJ Docs site, so that it will correctly feed analytics data into the CheerpJ GA4 property. 

I've included Claude's comment below so you've got sight of its summary: 

> Google Tag Manager (container GTM-WXJ325GT) is now installed on the CheerpJ docs site, and I verified it in a real build.
> 
> One note on placement: [src/content/docs]only holds markdown content — there's no <head> or <body> there. Those come from the shared theme package that all Leaning Tech sites use, which already had a GTM setup for browserpod. So I generalized that existing pattern rather than hardcoding a second copy:
> 
> - **New [gtm.ts]** a small per-product map of GTM container IDs: browserpod keeps its existing GTM-M4T2J9TS, and cheerpj3 (which is what every page of cheerpj.com resolves to, including all of /docs) gets your new GTM-WXJ325GT. Other sites (labs, cheerp.io, cheerpx.io) load no GT
> - **[BaseHead.astro:84-85]**The GTM script now uses the mapped container ID, and I moved it up so it renders immediately after <meta charset>, i.e. as high in <head> as possible per Google's instructions (it was previously near the bottom of the head).
> - **[Shell.astro:154-167]** — added the <noscript> iframe fallback immediately after the opening <body> tag. This part was missing entirely from the earlier browserpod install, so browserpod gains it too.
> Verification: I built the cheerpj site — all 80 real pages contain the script as the first element in <head> and the noscript right after <body>; the 128 files without it are all http-equiv="refresh" redirect stubs, which is expected. I also rebuilt the browserpod site to confirm the shared-theme change didn't disturb it: its 47 pages still carry GTM-M4T2J9TS and none picked up the CheerpJ container.
> 
> One thing to double-check on your side: the tag fires on all of cheerpj.com, not just /docs — that's how the site is built (the whole domain is one Astro site with base /docs/, plus the theme's shared pages). If you wanted it strictly limited to certain paths, that's better handled with a trigger condition inside the GTM container itself.

### DIFF
--- a/packages/astro-theme/components/BaseHead.astro
+++ b/packages/astro-theme/components/BaseHead.astro
@@ -2,6 +2,7 @@
 import { SITE_TITLE } from "../consts";
 import companyFavicon from "../assets/branding/company/tower.svg";
 import { productFromUrl } from "../lib/products";
+import { gtmContainerId } from "../lib/gtm";
 import { getLocalisedPaths } from "../lib/i18n";
 
 export interface Props {
@@ -75,11 +76,16 @@ if (isBlogPost && !plausibleDomains.includes("labs.leaningtech.com"))
 
 const localised = getLocalisedPaths(Astro.url.pathname, Astro.currentLocale);
 
+const gtmId = gtmContainerId(product);
 const gtmScript =
-	"(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-M4T2J9TS');";
+	gtmId &&
+	`(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','${gtmId}');`;
 ---
 
 <meta charset="utf-8" />
+
+<!-- Google Tag Manager -->
+{gtmScript && <script is:inline set:html={gtmScript} />}
 
 <link rel="preconnect" href="https://fonts.googleapis.com" />
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
@@ -131,9 +137,6 @@ const gtmScript =
 		<link rel="alternate" hreflang={lang} href={url} />
 	))
 }
-
-<!-- Google Tag Manager -->
-{product?.id === "browserpod" && <script is:inline set:html={gtmScript} />}
 
 <!-- Plausible analytics -->
 <script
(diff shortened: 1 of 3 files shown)


Liamoscarmccaffrey:
Passes Linter and Prettier.

Liamoscarmccaffrey:
@GabrielaReyna just a note about <pnpm-workspace.yaml>, it's added the following: 

```
allowBuilds:
  esbuild: set this to true or false
  sharp: set this to true or false
```
This may have just been to ensure it passes local build. But let me know if it needs to be removed and I'll take it out.

pnpm-workspace.yaml
```diff
@@ -2,6 +2,10 @@ packages:
   - sites/*
   - packages/*
 
+allowBuilds:
+  esbuild: set this to true or false
+  sharp: set this to true or false
+
```
codingfrog27:
```suggestion

```

Like you mentioned this should be removed. I think local build should pass without it as well and it'll cause the build to fail if we ever update to pnpm 11.

packages/astro-theme/components/BaseHead.astro
```diff
@@ -2,6 +2,7 @@
 import { SITE_TITLE } from "../consts";
 import companyFavicon from "../assets/branding/company/tower.svg";
 import { productFromUrl } from "../lib/products";
+import { gtmContainerId } from "../lib/gtm";
```
codingfrog27:
```suggestion
```

only needs to be imorted in shell.astro with new setup

packages/astro-theme/components/BaseHead.astro
codingfrog27, line 12:
```suggestion
gtmId?: string | undefined;
}
```
add id as optional prop that can be passed from parent (which in our case will always be passed by shell)

codingfrog27, line 47:
```suggestion
	image = product?.ogImage ?? "/social/common@2x.png",
	gtmId,
```

add it to props object

packages/astro-theme/components/BaseHead.astro
```diff
@@ -75,12 +76,17 @@ if (isBlogPost && !plausibleDomains.includes("labs.leaningtech.com"))
 
 const localised = getLocalisedPaths(Astro.url.pathname, Astro.currentLocale);
 
+const gtmId = gtmContainerId(product);
```
codingfrog27:
```suggestion
```

can be removed now since we inherit it from shell

packages/astro-theme/layouts/Shell.astro
codingfrog27, line 26:
```suggestion
			{image}
			{gtmId}
```
we're passing the id to basehead here

codingfrog27 (changes requested):
Hi Liam! Overall looks good, nice to have the clunky hardcoded product check removed :)

### Testing done
- checked pages and confirmed they had the container tag
- checked that browserpod pages kept its own tag

### Requested changes

- Like you already mentioned in your comment I think we should indeed get rid of the lines in the yaml file. They could break the build in the future.

- Also since basehead should always be a subcomponent of shell, we can actually just pass the value from shell into basehead, and we don't have to define it twice. Not a big issue but thought this way would be more streamlined, and less prone to bugs (since before you get the id out of potentially different urls), I've requested changes for this.

### Extra checks
- I assume we have the container itself live and published from the webUI right? if you're not getting data yet that's the first thing I'd double check
- I noticed we don't have anything for  consent management, think it's the same for browserpod and depending on what you want the gtm for it's also not needed, just thought i'd flag it.
- Be aware that this only adds tracking to the CJ docs and not the CJ site itself

Liamoscarmccaffrey:
@codingfrog27 happy to merge this if you are?

codingfrog27 (approved):
changes implemented

PR #440, opened 2026-07-30
Bp doc styling updates

smclacke:
**what I did**
- updated font with Stack Sans Text
- updated colours from greens to blues from provided colour scheme
- updated logo to full white matching browserpod.io 

 **to do / consider**
What is the plan with the logo, are we fully removing the green one? will we use white and black depending on light and dark mode and/or background where the logo is?
still green:
- favicon for both browserpod.io and browserpod.io/docs
- leaningtech github organisation readme 
- there are a few screenshots (bp-quickstart.png and express.png) that have the green logo


GabrielaReyna (changes requested):
Hi Sarah,

Great job with the restyling, I like it :)

**Replying to a few of your comments:**
- The docs don't have dark and light mode (yet) so you don't need to worry about the logo being white. 
- The green logo favicon needs to be swap for the white one from browserpod.io source-code. For the rest of green logo appearances, we need to map them and replace with white/black logo, we can do all of these this in a separate PR (s) after mapping them. 

**To-do in this PR:**
One thing to add that I forgot to mention but worth doing in this PR, is to _match the colour scheme of the code blocks to the same one used in browserpod.io_ (see the code editor component in the landing page at https://browserpod.io/). 

The colours provided by Alexis are #4a7dff (blue), #b7cdff (light blue), #020914(some kind of black), #c773da6 (intense pink), #ffd633 (yellow), #ff6161 (coral red). 

One thing to have in mind while doing this, is that the code-theme in the docs and the code-theme in browserpod's code editor component are implemented completely different:
- Docs code-block theme is set with Shiki (via astro-expressive-code), configured in packages/astro-theme/index.ts)
- Code editor theme in browserpod.io is CodeMirror + Lezer tags

Hence there is no direct copy-paste, you will need to work with what we currently use in the Docs, and change what we load on `loadShikiTheme("material-theme-darker")` for a custom theme with our colours. 

Just be careful that this theme is only for the BrowserPod's docs and not for the other docs sites. You might need to do something like this: 

> Add a theme field to Options in `packages/astro-theme/index.ts`, move the `loadShikiTheme(...)` call inside the integration function so it's parameterized per call, then have `sites/browserpod/astro.config.mjs` pass the custom palette while other sites keep passing "`material-theme-darker`". 

One thing to corroborate when setting up the custom colours is to check that the palette provided by Alexis and the colours currently used in browserpod.io are actually the same, there might be some variations here and there, just choose as you feel like it and looks better.

sites/browserpod/browserpod-code-theme.jsonc
```diff
@@ -0,0 +1,58 @@
+{
+	"name": "browserpod",
+	"type": "dark",
+	"colors": {
+		"editor.background": "#0d0d0e",
```
GabrielaReyna: This line might not be needed since we enforce a transparent background. We can just delete this line.

packages/astro-theme/index.ts
```diff
@@ -42,7 +33,23 @@ export default function ThemeIntegration(
...
 		hooks: {
-			"astro:config:setup": (params) => {
+			"astro:config:setup": async (params) => {
+				const theme =
+					typeof options.theme === "string" || options.theme === undefined
+						? await loadShikiTheme(options.theme ?? "material-theme-darker")
+						: options.theme;
+				theme.styleOverrides.frames = {
+					editorBackground: "transparent",
+					codeBackground: "transparent",
```
GabrielaReyna: I think this line is redundant

packages/astro-theme/index.ts
GabrielaReyna: I think these two lines might be doing nothing, please check it out and if this is dead, just remove

GabrielaReyna (changes requested):
Great work on this!

Let's fix the line jump issue in  in the small fixes PR you are working on. 

I added some comments on specific lines of code where it might be redundant to set a background colour or set transparency again when it is enforced already globally. 

Please check that out and test that nothing breaks if we get those removed.

GabrielaReyna (approved):

PR #446, opened 2026-08-11
Bp docs restyling

smclacke:
**Done:**

- Grey palette matched to the homepage
- Icons switched to mingcute for BP only
- Demos section removed from sidebar, dead links checked, `_headers` cleaned up
- Nav bar restyled to match the newer floating/glass reference: shape, size, colours, liquid-glass distortion effect

**To Do:**

- Slight crop at the very bottom of the page when overscrolling (sticky positioning limitation) need to change to fixed position but I got tons of bugs doing this that I don't have time for before release, something to return to if we want it to be exactly like the homepage


packages/astro-theme/components/nav/product/Nav.astro
```diff
@@ -129,4 +185,93 @@ let showcaseref = `https://labs.leaningtech.com/showcase${filterParam}`;
...
+
+	:global(.site-browserpod #product-nav site-search button) {
+		font-size: 14px;
+	}
+
+	:global(.site-browserpod) #product-nav .glass-filter {
+		position: absolute;
+		inset: 0;
+		z-index: 0;
+		backdrop-filter: blur(1px);
```
GabrielaReyna:
```suggestion
		backdrop-filter: blur(4px);
```

packages/astro-theme/components/nav/product/Nav.astro
```diff
@@ -129,4 +185,93 @@ let showcaseref = `https://labs.leaningtech.com/showcase${filterParam}`;
...
+		-webkit-backdrop-filter: blur(1px);
```
GabrielaReyna:
```suggestion
		-webkit-backdrop-filter: blur(4px);
```

packages/astro-theme/components/nav/product/Nav.astro
```diff
@@ -129,4 +185,93 @@ let showcaseref = `https://labs.leaningtech.com/showcase${filterParam}`;
...
+		pointer-events: none;
+	}
+
+	:global(.site-browserpod) #product-nav .glass-tint {
+		position: absolute;
+		inset: 0;
+		z-index: 1;
+		background-color: #d7dadd;
+		opacity: 0.06;
```
GabrielaReyna:
```suggestion
		opacity: 0.12;
```

GabrielaReyna (changes requested):
Hi Sarah!

Good job on this, some specific adjustments on the design:
- Please make the navbar blur effect stronger, I have suggested the same settings used for the browserpod site. Please test how it looks.
- Please remove the green dot animation next to the discord icon (this applies to all sites)


I am not still fully convinced on the width of the bar, specially when on big screens but I can't test it right now. We can leave this adjustment for the next iteration.

GabrielaReyna (changes requested):
Hi Sarah

Great work with this. I am just concerned about showing the online discord members without the green dot might confused users with the total number of discord members. 

I asked you to remove the green dot animation because it was resource consuming, but we have two options here:
- Keep a simple green dot next to the number of online members, but not animated.
- Remove the number of members and just keep the discord icon on its own (which matches browserpod.io)

Please try em both and go with what you think looks best.

More than that this should be ready to merge

Thank you!

smclacke:
Sorry, I didn't read your last comment correctly and understood 'remove green dot'. The animation has been removed and the star is now yellow :)

GabrielaReyna (approved):

PR #454, opened 2026-08-18
Feat: Created distribution version of Rust blog for sending to dev bl…

Liamoscarmccaffrey:
### Rationale
We're submitting the Rust post to developer community resources, for example This Week In Rust. Their contribution rules exclude submissions "of a commercial/sales nature". The purpose of this version of the blog is to retain the value of the walkthrough and the genuine innovation, while evading submission guidelines that reject excessively commercial content.


yuri91:
can you highlight what are the differences between the two versions? Which parts of the original post could be seen as commercial in nature?

Also I don't understand the changes on the original file (set to draft:true, some wording changes in the text, the banner also changed slightly). Are those intentional?

alexp-sssup:
@yuri91 Me and Liam agreed not go ahead with this changes at least for now

PR #455, opened 2026-08-19
CI cleanup

codingfrog27:
# CI cleanup 
(with GHA I'm refrring to Github Actions)
## Version updating
- **Node 20 → 22.23.2.** 
    - GHA was on 20 (which is EOL.
    - Set in package.json in `devEngines.runtime` which gets read by local, GHA and CF.
- **Actions bumped**: `checkout`/`setup-node` v3 → v7, `cache` v3 → v6,
  `pnpm/action-setup` v2 → v6, `rsync-deployments` 7.0.1 → v9 
- **pnpm pinned to 10.34.5** via `packageManager`. It was `version: 10`, a floating
  range that resolved to whatever was newest on the day the job ran.
- **caniuse-lite** → 1.0.30001810, clears the "browsers data is 9 months old" warning on every build.
- **Added allowBuilds** alongside the existing onlyBuiltDependencies, since pnpm renamed the setting in 10.26. Both kept for now.

## Removals
- **old deprecated navbar**: And the duplicate toolchains deprecated by its removal: (Vite 5, svelte 4 and esbuild 0.21.5)
- a deprecated monorepo branch deploy trigger.
- `.npmrc`: pnpm 11+ only read the auth/registry keys from it and other rules were ignored. Moved frozenlockfile rule to pnpm-workspace.yml and i dropped `auto-install-peers` since it's on by default. 

## Vale
- Fixed and re-enabled it.
-  Newer version also had different config so adjusted accordingly (deny.txt -> reject.txt)
- added exceptions for mdx imports, URLS's, code blocks and filenames.
- Turned off spelling since it flagged too many normal tech terms, now only checks capitalisation of our products etc. Also only checks files changed

## optimisations
- concurrency control for both workflows 
- linter only runs after PR is opened.
- links.yml crawled for cheerpj.io now cheerpj.com

# Todo after merge:

### Add dependabot for auto updating
- runs every x amount of time (set it to monthly) and opens PR for version bumps automatically with release notes in the description. 
- (also security PR's for when depends have published vulnaribilities)

### depend major version update/migration
- for astro, ts, canius and svelte
###  fix _redirects
- has 2 duplicate rules ignored at line 120 and 125 for CJ paths
- add the redirects for the changed BP doc urls

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,14 +10,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v7
 
       - name: Install Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
           node-version: 20
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v6
         name: Install pnpm
         with:
           version: 10
@@ -28,7 +28,7 @@ jobs:
         run: |
           echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v6
         name: Setup pnpm cache
         with:
           path: ${{ env.STORE_PATH }}
@@ -50,14 +50,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v7
 
       - name: Install Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
           node-version: 20
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v6
         name: Install pnpm
         with:
           version: 10
@@ -68,7 +68,7 @@ jobs:
         run: |
           echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v6
         name: Setup pnpm cache
         with:
           path: ${{ env.STORE_PATH }}
@@ -88,7 +88,7 @@ jobs:
     # temporarily disabled because test is brittle: https://github.com/errata-ai/vale/issues/841
     if: github.event_name == 'pull_request' && false
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v7
       - uses: errata-ai/vale-action@reviewdog
         with:
           version: 2.30.0
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,6 +2,10 @@ packages:
   - sites/*
   - packages/*
 
+allowBuilds:
+  esbuild: true
+  sharp: true
+
 onlyBuiltDependencies:
   - esbuild
   - sharp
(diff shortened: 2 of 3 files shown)


pnpm-workspace.yaml
GabrielaReyna, line 1: I think this file is making prettier fail, please format it.

.github/workflows/lint.yml
```diff
@@ -2,33 +2,38 @@ name: Lint
...
 
       - name: Install Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
-          node-version: 20
+          node-version-file: package.json
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v6
```
GabrielaReyna, line 26: Nothing to change here as doesn't affect us directly, just for our awareness that this action has an ongoing bug: https://github.com/pnpm/action-setup/issues/227

.vale.ini
```diff
@@ -1,13 +1,19 @@
...
 BasedOnStyles = Vale
+
+# Off: flagged 1400 ordinary tech terms as misspellings. Vale.Terms still
+# enforces brand capitalisation, Vale.Avoid still blocks reject.txt.
+Vale.Spelling = NO
+
+# Vale has no native MDX parser, so imports, URLs, code spans and filenames
+# all reach it as prose. Regexes must be bare; wrapping them in () breaks them.
+BlockIgnores = (?m)^import .*$
+TokenIgnores = https?://\S+, `[^`]+`, \S+\.(?:wasm|js|mjs|ts|json|astro|svelte|html|css|toml|yml)
```
GabrielaReyna: There is a tricky bug here that we need to be careful about, so using the vale warnings is actually doing its job.

Here we are choosing to ignore any string within codelines in backticks. The problem is that, when this happen, any appearances of the same word(s) before the code ticks will be dropped, even if they are an error. 

To avoid this we can change the middle items on `TokenIgnores`:

```
TokenIgnores = ..., `[^`]+`, ...
```
for 
```
TokenIgnores = ..., (`[^`]+`), ...
```

pnpm-workspace.yaml
```diff
@@ -2,6 +2,17 @@ packages:
...
+
+allowBuilds:
+  esbuild: true
+  sharp: true
+
 onlyBuiltDependencies:
   - esbuild
   - sharp
+# ^Only here as fallback: Corepack gives Cloudflare the
+# packageManager pnpm (>= 10.26.0), which reads allowBuilds.
```
GabrielaReyna: I got a bit confused given this comment said the opposite thing in a previous commit. 

Please just check on the Cloudflare dashboard the PNPM_VERSION setting (or equivalent) for each project and confirm it's actually ≥ 10.26.0 (or that Corepack really is overriding it), so we are sure that this comment is claiming something true.

GabrielaReyna (changes requested):
Hi Merlin!

Very clean work, thank you. Just added some minor comments for you to consider. 

As an additional note: let's make sure to remove the global navbar for webvm as it will give us some issues if the site is rebuilt after this PR is merged.

.github/workflows/lint.yml
codingfrog27: noted it!

.vale.ini
codingfrog27: ah good catch! implemented it.

pnpm-workspace.yaml
codingfrog27: yes fair! I got confused myself a bit but I've checked the logs.

There's no PNPM_VERSION env var in cloudflare and it's own default is 10.11.1. But the logs state: `Detected the following tools from environment: pnpm@10.34.5, nodejs@22.16.0
Preparing pnpm@10.34.5 for immediate activation...` .

So in other words Cloudflare reads the repo's packagemanager and uses its version, making the versions repo-controlled. 

I've updated the comment to hopefully make it a bit clearer. :)

PR #456, opened 2026-08-21
cx docs dataDevice writeFile corrections

smclacke:
Removed reference to Uint8Array in DataDevice writeFile doc since I can only find this function being passed a String.
Also fixed typo in writeFile file name and added this to redirects (specifically to not bother the user who flagged the issue in the first place / other users using that link on discord)

### DIFF
--- a/sites/cheerpx/public/_redirects
+++ b/sites/cheerpx/public/_redirects
@@ -2,3 +2,5 @@
 /blog        https://labs.leaningtech.com/blog        301
 /docs/showcase  https://labs.leaningtech.com/showcase  301
 /docs/community   https://labs.leaningtech.com/showcase    301
+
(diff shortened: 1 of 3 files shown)


coder0107git:
@smclacke @alexp-sssup Hey, sorry for the ping. The redirect added here breaks things (aka now there's a lovely 404). The file in question is named `writeFIle.md` not `writeFile.md`. Because of that the redirect now forces users to a page that doesn't exist. I'm working on PR that hopefully should fix this issue and a few others I've spotted.

smclacke:
hey @coder0107git thanks for letting us know. That's my mistake, I renamed the file and then reverted some things, also affecting that. I'll fix it again, the writeFIle.md is a typo of course.

PR #462, opened 2026-08-26
Cx docs - clarify mount point parent requirement

### LINKED ISSUE #459: Unclear expectations for CheerpX mounts
If you try to write the following, you get an error.

```js
const rootDevice = await CheerpX.DataDevice.create();
const webDevice = await CheerpX.WebDevice.create("/bin");

// Error!
const cx = await CheerpX.Linux.create({
  mounts: [
    { type: "dir", path: "/", dev: jsWriteOnlyFS },
    { type: "dir", path: "/app/bin", dev: webDevice },
  ],
});
```

If you know that this method expects the mount points to follow standard Linux conventions, that is, all mounts are attached to a parent in the hierarchy, not just free floating, then this behavior makes sense. The thing is though, this expectation isn't documented anywhere.

In other places where users mount their own file systems, like [ZenFS](https://github.com/zen-fs/core), you can simply do the following:

> ```js
> import { configure, InMemory } from "@zenfs/core";
> import { IndexedDB } from "@zenfs/dom";
> import { Zip } from "@zenfs/archives";
>
> const res = await fetch("mydata.zip");
>
> await configure({
>   mounts: {
>     "/mnt/zip": { backend: Zip, data: await res.arrayBuffer() },
>     "/tmp": InMemory,
>     "/home": IndexedDB,
>   },
> });
> ```

It would be nice to have the expectation of the Linux convention for the mounts to be documented better.

smclacke:
Closes #459 

Documents that every mount other than root needs an already-existing parent, following standard Linux mount conventions. Adds a warning callout to the `CheerpX.Linux.create` reference page with an example showing the failure case.

### DIFF
--- a/sites/cheerpx/src/content/docs/13-reference/00-CheerpX.Linux/00-create.md
+++ b/sites/cheerpx/src/content/docs/13-reference/00-CheerpX.Linux/00-create.md
@@ -47,7 +47,9 @@ interface MountPointConfiguration {
 	// 'proc' for process info files. (no device required)
 	type: "ext2" | "dir" | "devs" | "proc";
 
-	// First mount must be "/" (root)
+	// First mount must be "/" (root). Every other mount's path must
+	// have a parent that already exists, either from an earlier mount
+	// or a directory inside one, following standard Linux mount conventions.
 	path: string;
 	// Required for 'ext2' and 'dir' types, but optional for 'devs' and 'proc'
 	dev?: CheerpX.Device;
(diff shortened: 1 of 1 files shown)


sites/cheerpx/src/content/docs/13-reference/00-CheerpX.Linux/00-create.md
```diff
@@ -47,7 +47,9 @@ interface MountPointConfiguration {
 	// 'proc' for process info files. (no device required)
 	type: "ext2" | "dir" | "devs" | "proc";
 
-	// First mount must be "/" (root)
+	// First mount must be "/" (root). Every other mount's path must
+	// have a parent that already exists, either from an earlier mount
```
alexp-sssup: Well, it must always be from a directory inside a mount point, otherwise we would be overriding the mount point itself. Consider rephrasing this.

sites/cheerpx/src/content/docs/13-reference/00-CheerpX.Linux/00-create.md
```diff
@@ -67,6 +69,19 @@ const cx = await CheerpX.Linux.create({
 });
 ```
 
+> [!warning] Mount order matters
+> Because each mount's parent must already exist, mounts can't be "free floating". Mounting /app/bin before /app exists will fail, either mount /app first, or create that directory inside its parent filesystem before mounting a device on top of it.
```
alexp-sssup: > or create that directory inside its parent filesystem before mounting a device on top of it.

I don't actually understand this point

sites/cheerpx/src/content/docs/13-reference/00-CheerpX.Linux/00-create.md
```diff
@@ -67,6 +69,19 @@ const cx = await CheerpX.Linux.create({
...
+>
+> ```js
+> // Fails: "/app" doesn't exist yet, so "/app/bin" has no parent to mount onto
```
alexp-sssup: I think this not a good example since `/app` might exist in the ext2 image.

alexp-sssup (changes requested):

alexp-sssup (approved):

PR #463, opened 2026-08-26
Fix docs canonical URLs and add missing redirects

codingfrog27:
# Google search console report

This PR was prompted by google search engine reported 112 pages not being indexed. Luckily, upon further inspection this seemed a lot worse than it is. Only 7 pages are real content Google declined to index. Given this and the monorepo merge being imminent, I'm now switching back to the CI cleanup. Since a lot of these issues will be easier to fix with it being in one repo. That said, I do still have a couple small fixes that should take out some entries and give small qol improvements.

## The canonical URL bug

`packages/astro-theme/index.ts` sets `build: { format: "file" }`, so Astro emits `overview.html` and `Astro.url.pathname` carries the suffix throughout the build. Which in itself is not an issue since CF serves the URL extentionless and redirects the .html version to it.

But we built the canonical and og:url from the Astro pathname, resulting in every doc page declaring a canonical that points to a URL that points back to itself 😅.

This is the only place where we have links with .html on it (internal links and sitemap are all extensionless). This caused 24 "page with redirect" entries.

- The twitter url also used the .html version so I've changed that as well. :ok_hand: 

### Fix
Made canonical fallback and og:url take the `normalisedpath` which is extensionless. Note that this is a theme change that affects all doc sites.

## Sitemap
The sitemap had /docs hard coded on high priority. But /docs doesn't serve the homepage, only a meta refresh tag. **This is what causes the split second blank page.** Since this redirect happens at html level google won't index it. So it doesn't make sense to submit it as our highest priority URL. I've removed the entry, and will add a redirect after the monorepo merge.

# Post monorepo merge work

## _redirects
There were also 8 doc paths that have been changed since being indexed, added them to the `_redirects` file at first, but due to the dual deployment setup and differing roots and deployment rules between testing and production I'm saving this for after the merge.

I've documented them as well but for visibility's sake the redirects will be:
/docs/tutorials/expressjs   →  /docs/getting-started/expressjs
/docs/tutorials             →  /docs/getting-started
/docs/licensing             →  /docs/more/licensing
/docs/guides/nginx          →  /docs/guides/hosting
/docs/guides/natives        →  /docs/guides/working-around-native-npm-dependencies
/docs/demos/PackagePod      →  /showcase/package-pod/
/docs/demos/SaySomething    →  /showcase/say-something/
/docs/demos                 →  /showcase/
/docs/showcase              →  /showcase/
/docs/community             →  /showcase/
/docs/blog                  →  https://labs.leaningtech.com/blog
/docs  and  /docs/          →  /docs/overview          # replaces the meta-refresh stub

### Why add redirect rather than fixing our source?
Because there's no link to edit. The deprecated paths are the result of file renaming and moving. And the reference to the old URL's aren't in our repo but in googles index (plus potentially user bookmarks etc.).

### DIFF
--- a/packages/astro-theme/components/BaseHead.astro
+++ b/packages/astro-theme/components/BaseHead.astro
@@ -36,7 +36,7 @@ const browserpodCanonicals: Record<string, string> = {
 const normalisedPath = Astro.url.pathname.replace(/(?:\.html|\/)$/, "");
 const canonicalURL = browserpodCanonicals[normalisedPath]
 	? new URL(browserpodCanonicals[normalisedPath])
-	: new URL(Astro.url.pathname, canonicalSite);
+	: new URL(normalisedPath || "/", canonicalSite);
 
 const product = productFromUrl(canonicalURL);
 const favicon = product?.favicon ?? companyFavicon.src;
@@ -118,7 +118,7 @@ const gtmScript =
 
 <!-- Open Graph / Facebook -->
 <meta property="og:type" content="website" />
-<meta property="og:url" content={Astro.url} />
+<meta property="og:url" content={canonicalURL} />
 <meta property="og:title" content={title} />
 <meta property="og:description" content={description} />
 <meta property="og:image" content={new URL(image, Astro.url)} />
@@ -126,7 +126,7 @@ const gtmScript =
 
 <!-- Twitter -->
 <meta property="twitter:card" content="summary_large_image" />
-<meta property="twitter:url" content={Astro.url} />
+<meta property="twitter:url" content={canonicalURL} />
 <meta property="twitter:title" content={title} />
 <meta property="twitter:description" content={description} />
 <meta property="twitter:image" content={new URL(image, Astro.url)} />
--- a/sites/browserpod/src/pages/sitemap.xml.ts
+++ b/sites/browserpod/src/pages/sitemap.xml.ts
@@ -19,21 +19,13 @@ export const GET: APIRoute = async () => {
 		return true;
 	});
 
-	// The /docs/ landing page is served by the theme's pages/docs/index.astro,
-	// not the content collection, so it isn't in `entries`. Add it explicitly.
-	const landingUrl = `${BASE}/docs/`;
-
+	// The /docs/ landing page is served by the theme's pages/docs/index.astro, and it's only an html redirect, so it's not in the nav.
 	// entry.href is already "/docs/<slug>". Normalise and drop duplicates.
 	const innerUrls = Array.from(
 		new Set(entries.map((entry) => `${BASE}${entry.href}`))
 	).sort();
 
-	const pages = [
-		{ loc: landingUrl, priority: "0.8" },
-		...innerUrls
-			.filter((u) => u !== landingUrl)
-			.map((loc) => ({ loc, priority: "0.7" })),
-	];
+	const pages = innerUrls.map((loc) => ({ loc, priority: "0.7" }));
 
 	const xml = `<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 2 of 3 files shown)


sites/browserpod/src/pages/sitemap.xml.ts
```diff
@@ -19,21 +19,19 @@ export const GET: APIRoute = async () => {
...
 	).sort();
 
-	const pages = [
-		{ loc: landingUrl, priority: "0.8" },
-		...innerUrls
-			.filter((u) => u !== landingUrl)
-			.map((loc) => ({ loc, priority: "0.7" })),
-	];
+	const pages = innerUrls
+		.filter((u) => u !== landingUrl)
```
GabrielaReyna: I am a bit concerned about this filter, in my understanding,  this is pretty much dead code. 

`/docs/` was previously added manually (which is problematic with the google indexing because is a non existent page) and I think the filter meant to avoid /docs/ to be duplicated.  In this fix /docs/ is never added to the entries (which is correct, because it doesn't exist as a page) but then it is filtered out (which is wrong because it doesn't exist there in the first place, so it was dead code even. previous to this PR). 

In the escenario of actually having an index page (a /docs/ page) we would probably want it indexed anyway, so I think it is worth considering removing this line.

sites/browserpod/src/pages/sitemap.xml.ts
```diff
@@ -19,21 +19,19 @@ export const GET: APIRoute = async () => {
 		return true;
 	});
 
-	// The /docs/ landing page is served by the theme's pages/docs/index.astro,
-	// not the content collection, so it isn't in `entries`. Add it explicitly.
+	// The /docs/ landing page is deliberately absent. It is a meta-refresh stub
+	// that now 301s to /docs/overview via public/_redirects, so submitting it
```
GabrielaReyna: This comment isn't accurate and should be fixed to avoid any confusion:
- Is is **correct** to say `/docs/` is excluded from the sitemap as it has no real content of its own, it is not a page, but a meta stub . 
- It is **wrong** that the the comment says this happens "via public/_redirects" with a 301, which not true today. Right now `/docs/ ` is served by `pages/docs/index.astro`, which returns `200 OK` with a client-side meta-refresh to `/docs/overview`, not an HTTP redirect. The 301-via-_redirects version is planned post-merge, but isn't in place yet. Please update the comment to describe the current mechanism (or mark the 301 explicitly as a future/planned state) so it doesn't assert something the code doesn't actually do.

packages/astro-theme/components/BaseHead.astro
GabrielaReyna, line 1: I think the changes in this file are correct and safe right now, but let's keep a mental note of this for when the repo merging happens as this affects the sites at a theme level.

GabrielaReyna (changes requested):
Thanks for working on this Merlin. I left a couple of comments, please address them and this should be ready to merge after that.

sites/browserpod/src/pages/sitemap.xml.ts
codingfrog27: Ah yes you're right good catch! The comment and filter were still leftover from when I still had the implementations of the redirects in mind. Just removing the landingurl entirely.

Small note that if we would add an index page that I don't think the filter would catch it (if im reading it right) so we'd have to add it manually. But unless we'd specifically want an index landing page I think a redirect to the overview page would be easier anyways.

sites/browserpod/src/pages/sitemap.xml.ts
codingfrog27: Ah yes same here as for the above comment, amending this too👌

GabrielaReyna (approved):

